PR #21, opened 2017-05-20
update Vert.x MQTT server component version in docs

boliza:

PR #22, opened 2017-05-20
changed version of artifact 3.4.0 -> 3.4.1

Sammers21:
The problem is that maven can't find 3.4.0 version of artifact in the repository (https://oss.sonatype.org/content/repositories/snapshots). I have fixed it changing the version to 3.4.1 .


Sammers21:
I have noticed that #21 is the same issue and it was created an hour before. I'm sorry. Should the pull request be closed?

ppatierno:
Yes @Sammers21 I'm going to close this PR. The right way is to update the package-info and not the ascii doc files because they are generated automatically.

PR #29, opened 2017-07-11
Refactor net base for pipeline

vietj:


src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -85,8 +80,34 @@ public MqttServer listen(Handler<AsyncResult<MqttServer>> listenHandler) {
   public MqttServer listen(int port, String host, Handler<AsyncResult<MqttServer>> listenHandler) {
     Handler<MqttEndpoint> h1 = endpointHandler;
     Handler<Throwable> h2 = exceptionHandler;
-    Handler<MqttConnection> mqttConnectionHandler = c -> c.init(h1, h2);
-    listen(mqttConnectionHandler, port, host, ar -> listenHandler.handle(ar.map(this)));
+    server.connectHandler(so -> {
+      NetSocketInternal soi = (NetSocketInternal) so;
+      soi.messageHandler(o -> {});
```
ppatierno: why the need for setting an "empty" handler here when it will be set later (line 103) ?

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -85,8 +80,34 @@ public MqttServer listen(Handler<AsyncResult<MqttServer>> listenHandler) {
...
+      ChannelPipeline pipeline = soi.channelHandlerContext().pipeline();
+
+/*
+      if (logEnabled) {
+        pipeline.addLast("logging", new LoggingHandler());
+      }
+*/
```
ppatierno: why these lines are commented ?

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
vietj: it should be removed indeed

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
vietj: it should not be, I thought that NetSocket was doing the same internally, perhaps it should

vietj: it should not be removed, as the NetServerImpl already does it along with other base handlers

vietj:
updated according to the feedback

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -85,8 +80,27 @@ public MqttServer listen(Handler<AsyncResult<MqttServer>> listenHandler) {
   public MqttServer listen(int port, String host, Handler<AsyncResult<MqttServer>> listenHandler) {
     Handler<MqttEndpoint> h1 = endpointHandler;
     Handler<Throwable> h2 = exceptionHandler;
-    Handler<MqttConnection> mqttConnectionHandler = c -> c.init(h1, h2);
-    listen(mqttConnectionHandler, port, host, ar -> listenHandler.handle(ar.map(this)));
+    server.connectHandler(so -> {
+      NetSocketInternal soi = (NetSocketInternal) so;
+      ChannelPipeline pipeline = soi.channelHandlerContext().pipeline();
+      pipeline.addBefore("handler", "mqttEncoder", MqttEncoder.INSTANCE);
```
Sammers21: Should we have logging here? I mean something like:
```java

if (logEnabled) {
    pipeline.addLast("logging", new LoggingHandler());
}


```

ppatierno: It was moved in the NetServerImpl (line 138), no need to have it here.

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -102,35 +116,7 @@ public synchronized MqttServer exceptionHandler(Handler<Throwable> handler) {
...
-      pipeline.addLast("mqttDecoder", new MqttDecoder(this.mqttServerOptions.getMaxMessageSize()));
-    } else {
-      // max message size not set, so the default from Netty MQTT codec is used
-      pipeline.addLast("mqttDecoder", new MqttDecoder());
-    }
-  }
-
-  @Override
-  protected Object safeObject(Object msg, ByteBufAllocator allocator) {
+  private Object safeObject(Object msg, ByteBufAllocator allocator) {
```
ppatierno: I think that now we can get rid of this safeObject method because it's not inherited from a class. We can do the stuff inside the handleMessage of the MqttConnection. I'm going to merge and I already have the update for that to commit on the merged master.

PR #36, opened 2017-07-25
add handling of qos=1 and qos=2 message publishing

Sammers21:
Copy of the PR: https://github.com/vert-x3/vertx-mqtt-client/pull/8


src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -21,20 +21,7 @@
...
-import io.netty.handler.codec.mqtt.MqttFixedHeader;
-import io.netty.handler.codec.mqtt.MqttMessageFactory;
-import io.netty.handler.codec.mqtt.MqttMessageIdVariableHeader;
-import io.netty.handler.codec.mqtt.MqttMessageType;
-import io.netty.handler.codec.mqtt.MqttPublishVariableHeader;
-import io.netty.handler.codec.mqtt.MqttQoS;
-import io.netty.handler.codec.mqtt.MqttSubscribePayload;
-import io.netty.handler.codec.mqtt.MqttTopicSubscription;
-import io.netty.handler.codec.mqtt.MqttUnsubscribePayload;
+import io.netty.handler.codec.mqtt.*;
```
ppatierno: I think you are using Eclipse IDE right ? I'm using IntelliJ :-) Can we avoid the usage of "*" otherwise every time I change something and you do the same the packages appears and disappears :-)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -265,6 +261,15 @@ public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolea
...
     }
 
+    switch (qosLevel.value()) {
+      case 1:
+        qos1outbound.add(publish);
+        break;
+      case 2:
+        qos2outbound.add(publish);
+        break;
+    }
```
ppatierno: Why not using the MqttQos enum ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Nope, I'm using IntelliJ :) But I get your point, so I will avoid such things

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Sure, using the MqttQos enum is better

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -265,6 +279,15 @@ public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolea
       publishSentHandler.handle(Future.succeededFuture(variableHeader.messageId()));
     }
 
+    switch (qosLevel) {
```
Sammers21: Exactly

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -429,7 +452,7 @@ public MqttClient pingResponseHandler(Handler<Void> pingResponseHandler) {
   public MqttClient ping() {
 
     MqttFixedHeader fixedHeader =
-      new MqttFixedHeader(MqttMessageType.PINGREQ, false, MqttQoS.AT_MOST_ONCE, false, 0);
+      new MqttFixedHeader(MqttMessageType.PINGREQ, false, AT_MOST_ONCE, false, 0);
```
Sammers21: There are should not be any changes here. I have canceled these changes.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -500,19 +523,20 @@ void publishComplete(int publishMessageId) {
...
    * Sends the PUBREL message to server
    *
-   * @param publishMessageId  identifier of the PUBLISH message to acknowledge
+   * @param publishMessageId identifier of the PUBLISH message to acknowledge
    */
   void publishRelease(int publishMessageId) {
 
     MqttFixedHeader fixedHeader =
-      new MqttFixedHeader(MqttMessageType.PUBREL, false, MqttQoS.AT_LEAST_ONCE, false, 0);
+      new MqttFixedHeader(MqttMessageType.PUBREL, false, AT_LEAST_ONCE, false, 0);
```
Sammers21: There are should not be any changes here. I have canceled these changes.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -597,8 +621,12 @@ void handleUnsuback(int unsubackMessageId) {
   void handlePuback(int pubackMessageId) {
 
     synchronized (this.connection) {
-      if (this.publishCompleteHandler != null) {
-        this.publishCompleteHandler.handle(pubackMessageId);
+      if (((MqttPublishVariableHeader) qos1outbound.peek().variableHeader()).messageId() == pubackMessageId) {
```
ppatierno: here the first problem comes. The structure you need is something hybrid ... 
It works like a queue, because of course you need to publish in the order the user is calling the publish method but the ack could be received out of order so `peek()` and `poll()` don't fit well here. Imagine that you send two PUBLISH messages with id = 1 and id = 2 but the you receive the PUBACK for id = 2 first.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -611,8 +639,13 @@ void handlePuback(int pubackMessageId) {
   void handlePubcomp(int pubcompMessageId) {
 
     synchronized (this.connection) {
-      if (this.publishCompleteHandler != null) {
-        this.publishCompleteHandler.handle(pubcompMessageId);
+      if (((MqttPublishVariableHeader) qos2outbound.peek().variableHeader()).messageId() == pubcompMessageId) {
```
ppatierno: Ditto as for QoS 1

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Sure, I have just updated PR using LinkedHashMap

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -79,6 +81,15 @@ public MqttClientOptions(MqttClientOptions other) {
     super(other);
   }
 
+  public int getMaxInflightQueue() {
```
ppatierno: Summary for this method

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -79,6 +81,15 @@ public MqttClientOptions(MqttClientOptions other) {
...
+    return maxInflightQueue;
+  }
+
+  public MqttClientOptions setMaxInflightQueue(int maxInflightQueue) {
```
ppatierno: Summary for this method

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -538,6 +587,9 @@ void publishRelease(int publishMessageId) {
 
     io.netty.handler.codec.mqtt.MqttMessage pubrel = MqttMessageFactory.newMessage(fixedHeader, variableHeader, null);
 
+    // we need to remove here because we want to add the element to the end of list
+    qos2outbound.remove(publishMessageId);
```
ppatierno: not sure we need to change the order, putting the PUBREL at the end of the queue

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -668,7 +736,7 @@ void handlePubcomp(int pubcompMessageId) {
   void handlePubrec(int pubrecMessageId) {
 
     synchronized (this.connection) {
-      this.publishRelease(pubrecMessageId);
+        this.publishRelease(pubrecMessageId);
```
ppatierno: there is more than one tab ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -490,18 +537,20 @@ void publishAcknowledge(int publishMessageId) {
...
       new MqttFixedHeader(MqttMessageType.PUBREC, false, AT_MOST_ONCE, false, 0);
 
     MqttMessageIdVariableHeader variableHeader =
-      MqttMessageIdVariableHeader.from(publishMessageId);
+      MqttMessageIdVariableHeader.from(publishMessage.messageId());
 
     io.netty.handler.codec.mqtt.MqttMessage pubrec = MqttMessageFactory.newMessage(fixedHeader, variableHeader, null);
 
+    qos2inbound.put(publishMessage.messageId(), new CurrentUnacknowledgedAndPublishMessage(pubrec, publishMessage));
+    countInflightQueue++;
```
ppatierno: the inflight queue is intended to be only for outbound messages

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -727,10 +795,19 @@ void handlePublish(MqttPublishMessage msg) {
...
-      this.publishComplete(pubrelMessageId);
+      CurrentUnacknowledgedAndPublishMessage message = qos2inbound.get(pubrelMessageId);
+
+      if (message == null) {
+        return;
+      }
 
-      // TODO : call publishHandler with the message received in the PUBISH
-      //        we have to save it somewhere (maybe in a queue)
+      countInflightQueue--;
```
ppatierno: the inflight queue is intended only for outbound messages

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -490,18 +537,20 @@ void publishAcknowledge(int publishMessageId) {
...
     MqttFixedHeader fixedHeader =
       new MqttFixedHeader(MqttMessageType.PUBREC, false, AT_MOST_ONCE, false, 0);
 
     MqttMessageIdVariableHeader variableHeader =
-      MqttMessageIdVariableHeader.from(publishMessageId);
+      MqttMessageIdVariableHeader.from(publishMessage.messageId());
 
     io.netty.handler.codec.mqtt.MqttMessage pubrec = MqttMessageFactory.newMessage(fixedHeader, variableHeader, null);
 
+    qos2inbound.put(publishMessage.messageId(), new CurrentUnacknowledgedAndPublishMessage(pubrec, publishMessage));
```
ppatierno: My first impression is that we can just make it simpler saving the `publishMessage` as for the outbound queus.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -119,6 +134,20 @@ public MqttClientImpl(Vertx vertx, MqttClientOptions options) {
   }
 
   /**
+   * This class represents a pair of MqttMessage and MqttPublishMessage
+   * It uses in qos2inbound for calling publishHandler when PUBREL arrives
+   */
+  static class CurrentUnacknowledgedAndPublishMessage {
```
ppatierno: See following comments, this class could not be needed.

src/main/java/io/vertx/mqtt/MqttClientOptions.java

src/main/java/io/vertx/mqtt/MqttClientOptions.java

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Sure

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: You are right

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Sure

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: I m agree

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Yeah, fixed :-)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -32,6 +32,7 @@
   private static final String DEFAULT_HOST = "localhost";
   private static final int DEFAULT_WILL_QOS = 0;
   private static final int DEFAULT_KEEP_ALIVE_TIME_SECONDS = 30;
+  private static final int DEFAULT_MAX_INFLIGHT_QUEUE = 30;
```
Sammers21: we also should think about default value for maxInflightQueue

ppatierno: Let's do the Paho way, use 10

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -196,6 +198,23 @@ public MqttClientOptions setKeepAliveTimeSeconds(int keepAliveTimeSeconds) {
...
 
   /**
+   * @return max count of unacknowledged messages at the same time
+   */
+  public int getMaxInflightQueue() {
+    return maxInflightQueue;
+  }
+
+  /**
+   * Set max count of unacknowledged messages at the same time
```
ppatierno: please remove "at the same time" ;)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -640,6 +675,14 @@ void handleUnsuback(int unsubackMessageId) {
   void handlePuback(int pubackMessageId) {
 
     synchronized (this.connection) {
+      io.netty.handler.codec.mqtt.MqttMessage removedPacket = qos1outbound.remove(pubackMessageId);
+
+      if (removedPacket == null) {
+        return;
```
ppatierno: Add a log warning here, because a PUBACK is arrived but we don't have the corresponding message-id in the qos1outbound queue ... it's a weird situation :-)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -654,6 +697,14 @@ void handlePuback(int pubackMessageId) {
   void handlePubcomp(int pubcompMessageId) {
 
     synchronized (this.connection) {
+      io.netty.handler.codec.mqtt.MqttMessage removedPacket = qos2outbound.remove(pubcompMessageId);
+
+      if (removedPacket == null) {
+        return;
```
ppatierno: Ditto as above. Add a log warning here, because a PUBCOMP is arrived but we don't have the corresponding message-id in the qos2outbound queue ... it's a weird situation :-)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -727,10 +778,17 @@ void handlePublish(MqttPublishMessage msg) {
   void handlePubrel(int pubrelMessageId) {
 
     synchronized (this.connection) {
-      this.publishComplete(pubrelMessageId);
+      MqttMessage message = qos2inbound.get(pubrelMessageId);
+
+      if (message == null) {
+        return;
```
ppatierno: Ditto as above. Add a log warning here, because a PUBREL is arrived but we don't have the corresponding message-id in the qos2inbound queue ... it's a weird situation :-)

src/main/java/io/vertx/mqtt/MqttClientOptions.java

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -196,6 +198,23 @@ public MqttClientOptions setKeepAliveTimeSeconds(int keepAliveTimeSeconds) {
...
   /**
+   * @return max count of unacknowledged messages
+   */
+  public int getMaxInflightQueue() {
+    return maxInflightQueue;
+  }
+
+  /**
+   * Set max count of unacknowledged messages
+   * @param maxInflightQueue max count
```
ppatierno: add "... of unacknowledged messages"

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -100,9 +102,22 @@
   //handler to call when the remote MQTT server closes the connection
   Handler<Void> closeHandler;
 
+  // storage of PUBLISH QoS=1 messages which was not responded with PUBACK
+  LinkedHashMap<Integer, io.netty.handler.codec.mqtt.MqttMessage> qos1outbound = new LinkedHashMap<>();
+
+  // storage of of PUBLISH QoS=2 messages which was not responded with PUBREC
```
ppatierno: type double "of"

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -267,6 +289,17 @@ public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolea
...
+    switch (qosLevel) {
+      case AT_LEAST_ONCE:
+        qos1outbound.put(variableHeader.messageId(), publish);
+        break;
+      case EXACTLY_ONCE:
+        qos2outbound.put(variableHeader.messageId(), publish);
+        break;
+    }
+
+    countInflightQueue++;
```
ppatierno: incrementing here, we have that it counts the QoS = 0 message as well which won't go into any queue. It's not right to increment this counter here.

src/main/java/io/vertx/mqtt/MqttClientOptions.java

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -267,6 +289,17 @@ public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolea
...
+        break;
+      case EXACTLY_ONCE:
+        qos2outbound.put(variableHeader.messageId(), publish);
+        countInflightQueue++;
+        break;
+    }
+
     this.write(publish);
 
     if (publishSentHandler != null) {
```
ppatierno: Maybe the publishSentHandler should be called here only if QoS = 0. If Qos is 1 or 2, the handler should be called when PUBACK is received (QoS 1) and PUBCOMP is received (QoS 2). It means that the publish method should not have this handler as parameter but we would need an "ad hoc" method to set it (as it happens for other handlers). It's a breaking change on the API, I know. Wdyt ?

Sammers21: But, actually, we have publishCompleteHandler which called when PUBACK or PUBCOM is received. About publishSentHandler ... I think that it would better to just remove it from params... it seems that it is useless because it can be replaced with some code after the method call.

ppatierno: Sorry you are right. I forgot about that. For now I would not delete the publishSentHandler; it's useful to the user to know which is the message-id assigned to the publish so that he can do some check when the publishCompleteHandler is called.

Sammers21:
@ppatierno, I have just fixed the conflicts

vietj:
great, I'll make a general review of the API

Sammers21:
@ppatierno , I have just fixed that

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -251,6 +266,13 @@ public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolea
   @Override
   public MqttClient publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, Handler<AsyncResult<Integer>> publishSentHandler) {
 
+    if (countInflightQueue >= options.getMaxInflightQueue()) {
+      publishSentHandler.handle(Future.failedFuture(
+        String.format("Attempt to exceed the limit of %d inflight messages", options.getMaxInflightQueue())
+      ));
```
ppatierno: I have just added an `MqttException` class for handling "internal" errors. I have used it for invalid topic name and filter. You should add a new reason code there (i.e. `INFLIGHT_QUEUE_FULL`) and using such exception for the failed future.

Sammers21: Done

Sammers21:
@ppatierno , we also should have a similar test for the QoS 2 right?

Sammers21:
@ppatierno, I have just created some tests for QoS 1 and 2

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -0,0 +1,213 @@
...
+    expectOrder.add(3);
+
+
+    client.publishCompletionHandler(h -> {
+      context.assertTrue(h.intValue() == expectOrder.poll());
+      log.info("[CLIENT] Publish completed for message with id: " + h);
+      async.countDown();
+    });
+
+    client.connect(1883, "localhost", c -> {
```
ppatierno: Please use always constants for server port and hostname

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -0,0 +1,213 @@
...
```
ppatierno: Ditto as above

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    async.await();
+    client.disconnect();
+  }
+
+
+  @Before
+  public void before() {
+    MqttServerOptions serverOptions = new MqttServerOptions();
+    server = MqttServer.create(vertx, serverOptions);
```
ppatierno: You can use the other `create` overload with no options as parameter (currently you are passing a default options that is what the other overload does).

Sammers21: Sure

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
Sammers21: Ok

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
Sammers21: Done

Sammers21:
@ppatierno, you are right, better to initialize server inside before method. I have just updated the PR reducing duplicates in the test.

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -0,0 +1,160 @@
...
+    mqttEndpoint.publishHandler(p -> {
+      log.info("[SERVER] Received PUBLISH with message id = " + p.messageId());
+
+      if (p.qosLevel().equals(MqttQoS.EXACTLY_ONCE))
+        // when received last PUBLISH message to acknowledge
+        if (p.messageId() == 3) {
+          // send PUBREC in order 3 --> 2 --> 1 -->
+          mqttEndpoint.publishReceived(3);
+          mqttEndpoint.publishReceived(2);
+          mqttEndpoint.publishReceived(1);
```
ppatierno: why the order here is different and not 2,1 and 3 ?

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -0,0 +1,160 @@
...
+          mqttEndpoint.publishAcknowledge(1);
+          mqttEndpoint.publishAcknowledge(3);
+        }
+    });
+
+    mqttEndpoint.publishReleaseHandler(pr -> {
+      log.info("[SERVER] Receive PUBREL with message id = " + pr);
+
+      // when received last PUBREL message to acknowledge
+      if (pr == 1) {
```
ppatierno: Related to the previous question it should be `pr == 3` if we change to 2,1 and 3 the PUBREC order ?

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
Sammers21: So the PUBREC order here is not 2,1 and 3, but 3 2 1, however, the PUBCOMP order is 2,1, 3. I have changed the order two times during a publish session to make sure that client can receive PUBCOMP packets in a different order.

Sammers21: If we will have 2,1 and 3 here that means that client will receive PUBREC packets and PUBCOM packets in the same order

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
Sammers21: Sure

src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
ppatierno: So are you adding a more shuffling here ? It seems a good idea but please comment that in the code.

Sammers21: Sure. I put some comments for that in code.

Sammers21:
squashed

PR #38, opened 2017-07-25
add the opportunity to config max message size

Sammers21:


src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -230,6 +232,17 @@ public boolean isAutoGeneratedClientId() {
...
+    return maxMessageSize;
+  }
+
+  public MqttClientOptions setMaxMessageSize(int maxMessageSize) {
+    Arguments.require(maxMessageSize > 0 , "maxMessageSize must be > 0");
+    setReceiveBufferSize(maxMessageSize);
+    this.maxMessageSize = maxMessageSize;
+    return this;
+  }
+
```
ppatierno: Why not having the same logic as the MQTT server ?

Sammers21: Yes, I think it better to having the same logic

Sammers21: But I m just a bit disappointed by the fact we have two methods for limiting the incoming messages size: setMaxMessageSize and setReceiveBufferSize from extended class.

So in case, we want to limit incoming message size should we call two methods for that purpose?  Can we better call only one method instead?

Sammers21: I know that ReceiveBufferSize and MaxMessageSize are different things but in fact, MQTT client(as well as MQTT server) deal only with MQTT packets

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -230,6 +233,36 @@ public boolean isAutoGeneratedClientId() {
     return isAutoGeneratedClientId;
   }
 
+  public int getMaxMessageSize() {
```
ppatierno: Add a summary for this method

Sammers21:
@ppatierno, I have also added exceptionHandler to the API so for being able to catch DecodeExceprion

ppatierno:
@Sammers21 good point but it's better having two different PRs

Sammers21:
@ppatierno , sure. I have created [the PR](https://github.com/vert-x3/vertx-mqtt/pull/41) for that

src/test/java/io/vertx/mqtt/test/client/MqttClientMaxMessageSizeTest.java
```diff
@@ -0,0 +1,72 @@
...
+//   this functionality should be introduced in the PR https://github.com/vert-x3/vertx-mqtt/pull/41
+//
+//    client.exceptionHandler(t->{
+//      log.error("Exception raised", t);
+//
+//      if (t instanceof DecoderException) {
+//        log.info("PUBLISH <--- message with big size");
+//        async.countDown();
+//      }
+//    });
```
ppatierno: Does it mean that we have to merge the other PR first for having this merged right ?

Sammers21: Yes :-)

ppatierno: @Sammers21 I have merged that PR, now you can de-comment this lines in the test then squashing the commits and I'll merge it ;)

ppatierno:
@Sammers21 can you squash the commits ?

Sammers21:
@ppatierno, I have just squashed.

Issue #39, opened 2017-07-25
Add a closeHandler to the client


ppatierno:
It could be useful to have the possibility to pass an handler to the client which is called when the connection is closed. In MQTT it can happen that the server closes the connection for different reasons so it's important to have the client being notified about that.

ppatierno:
@Sammers21 can you work on that proposing a PR ?

Sammers21:
@ppatierno , sure

PR #40, opened 2017-07-25
add a closeHandler to the client

Sammers21:
Solution for #39 issue

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -95,6 +95,8 @@ public class MqttClientImpl implements MqttClient {
   Handler<AsyncResult<MqttConnAckMessage>> connectHandler;
   // handler to call when a pingresp is received
   Handler<Void> pingrespHandler;
+  //handler to call when the remote MQTT server closes the connection
+  Handler<Void> closeHandler;
 
   // counter for the message identifier
   private int messageIdCounter;
@@ -152,6 +154,7 @@ public class MqttClientImpl implements MqttClient {
         this.connection = new MqttClientConnection(this, soi, options);
 
         soi.messageHandler(msg -> connection.handleMessage(msg));
+        soi.closeHandler(v -> handleClosed());
 
         MqttFixedHeader fixedHeader = new MqttFixedHeader(MqttMessageType.CONNECT,
           false,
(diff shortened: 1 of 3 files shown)


Sammers21:
I have tried to do this in the same way as it in MQTT server. Is it ok now?

src/main/java/io/vertx/mqtt/impl/MqttClientConnection.java
```diff
@@ -21,6 +21,7 @@
 import io.netty.handler.codec.DecoderResult;
 import io.netty.handler.codec.mqtt.MqttMessage;
 import io.netty.handler.codec.mqtt.MqttMessageIdVariableHeader;
+import io.vertx.core.Handler;
```
ppatierno: It's not needed anymore right ?

src/main/java/io/vertx/mqtt/impl/MqttClientConnection.java
```diff
@@ -239,4 +240,5 @@ void close(){
   void writeMessage(MqttMessage mqttMessage) {
     so.writeMessage(mqttMessage);
   }
+
```
ppatierno: If you remove the above import and this blank line you can just revert this file to the original one because there are no changes.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -152,6 +154,7 @@ public MqttClient connect(Handler<AsyncResult<MqttConnAckMessage>> connectHandle
         this.connection = new MqttClientConnection(this, soi, options);
 
         soi.messageHandler(msg -> connection.handleMessage(msg));
+        soi.closeHandler((nothing) -> handleClosed());
```
ppatierno: It's better a simple "v" (stands for Void) than "(nothing)" :-)

src/main/java/io/vertx/mqtt/impl/MqttClientConnection.java
Sammers21: Sure

src/main/java/io/vertx/mqtt/impl/MqttClientConnection.java
Sammers21: Yeah

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Done :)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -152,6 +154,7 @@ public MqttClient connect(Handler<AsyncResult<MqttConnAckMessage>> connectHandle
         this.connection = new MqttClientConnection(this, soi, options);
 
         soi.messageHandler(msg -> connection.handleMessage(msg));
+        soi.closeHandler((v) -> handleClosed());
```
ppatierno: sorry to be pedantic ... one parameter no need for parenthesis :-) After that you can squash and I'll merge

Sammers21: Done

PR #44, opened 2017-08-05
minor refactoring + client SSL test

Sammers21:
Changes:
* MqttConnection was renamed to MqttServerConnection. I think it is better because we have MqttCleintConnection for a client.

* setTrustAll method in MqttClientOptions was overridden

* SSL test

### DIFF
--- a/src/main/java/io/vertx/mqtt/MqttClientOptions.java
+++ b/src/main/java/io/vertx/mqtt/MqttClientOptions.java
@@ -17,9 +17,16 @@
 package io.vertx.mqtt;
 
 import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.impl.Arguments;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.net.JksOptions;
+import io.vertx.core.net.KeyCertOptions;
 import io.vertx.core.net.NetClientOptions;
+import io.vertx.core.net.PemKeyCertOptions;
+import io.vertx.core.net.PemTrustOptions;
+import io.vertx.core.net.PfxOptions;
+import io.vertx.core.net.TrustOptions;
 
 /**
  * Represents options used by the MQTT client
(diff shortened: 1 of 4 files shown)


src/test/java/io/vertx/mqtt/test/client/MqttClientSslTest.java
```diff
@@ -0,0 +1,100 @@
...
+      .setPort(MQTT_SERVER_TLS_PORT)
+      .setHost(MQTT_SERVER_HOST)
+      .setKeyCertOptions(pemKeyCertOptions)
+      .setSsl(true);
+
+    MqttClientOptions clientOptions = new MqttClientOptions()
+      .setPort(MQTT_SERVER_TLS_PORT)
+      .setHost(MQTT_SERVER_HOST)
+      .setSsl(true)
+      .setTrustAll(true);
```
ppatierno: Can you add another test where you don't use "trust all" but use the client-truststore we have in the resources ? Just to confirm that the handshake is going well with the server certificate. Finally, did you take a look at the network traffic with Wireshark just to have another way to check the SSL/TLS handshake ?

Sammers21: image

Sammers21: this is what Wireshark shows when running just added "clientSslClientTruststoreTest" test

Sammers21: Btw, Wireshark is a great tool.

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -272,6 +273,17 @@ public MqttClientOptions setSsl(boolean ssl) {
...
+  public MqttClientOptions setTrustStoreOptions(JksOptions options) {
+    super.setTrustStoreOptions(options);
+    return this;
+  }
+
+  @Override
+  public MqttClientOptions setTrustAll(boolean trustAll) {
+    super.setTrustAll(trustAll);
+    return this;
+  }
```
ppatierno: As already provided by the server component, even if without tests, we should add even the following ways to provide certs (as PEM, PFX, ...) overriding following methods : 

`setKeyCertOptions`
`setKeyStoreOptions`
`setPfxKeyCertOptions`
`setPemKeyCertOptions`
`setTrustOptions`
`setPemTrustOptions`
`setPfxTrustOptions`

and even adding following override to be more complete : 

`addEnabledCipherSuite`
`addEnabledSecureTransportProtocol`
`addCrlPath`
`addCrlValue`

src/test/java/io/vertx/mqtt/test/client/MqttClientSslTest.java
ppatierno: I know, it's my preferred tool :-)

src/main/java/io/vertx/mqtt/MqttClientOptions.java

ppatierno:
@Sammers21 I'm good to go but you should squash :-)

Sammers21:
@ppatierno, squashed

ppatierno:
@Sammers21 thanks ! I'm going to merge it !

PR #52, opened 2017-08-30
Mark reactivex packages as optional in OSGi metadata

sauthieg:
When trying the `3.5.0.Beta1` bundle, it was unable to deploy when rxjava2 was missing.

I've added `resolution:=optional` to theses new imported packages (just like what was done for other languages).

I've also discovered that `io.vertx.ext.auth` was actually never used in this module, but import was required. I've simply removed the lines related to this group of packages.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,12 +196,16 @@
             io.vertx.ext.auth,\
             io.vertx.ext.auth.*;resolution:=optional,\
             io.vertx.lang.rxjava.*;resolution:=optional,\
+            io.vertx.lang.reactivex.*;resolution:=optional,\
             io.vertx.lang.groovy.*;resolution:=optional,\
             io.vertx.codegen.annotations;resolution:=optional,\
             io.vertx.rx.java;resolution:=optional,\
             io.vertx.rxjava.core.*;resolution:=optional,\
             io.vertx.rxjava.ext.auth.*;resolution:=optional,\
+            io.vertx.reactivex;resolution:=optional,\
+            io.vertx.reactivex.core.*;resolution:=optional,\
             rx.*;resolution:=optional,\
+            io.reactivex.*;resolution:=optional,\
             *
           -exportcontents: !*impl, !examples, *
           ]]></bnd>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -193,15 +193,17 @@
             org.codehaus.groovy.*;resolution:=optional,\
             kotlin.*;resolution:=optional,\
             io.vertx.groovy.*;resolution:=optional,\
-            io.vertx.ext.auth,\
-            io.vertx.ext.auth.*;resolution:=optional,\
```
cescoffier: Why are these removed and not io.vertx.rxjava.ext.auth ?

pom.xml
```diff
@@ -193,15 +193,17 @@
...
             io.vertx.lang.rxjava.*;resolution:=optional,\
+            io.vertx.lang.reactivex.*;resolution:=optional,\
             io.vertx.lang.groovy.*;resolution:=optional,\
             io.vertx.codegen.annotations;resolution:=optional,\
             io.vertx.rx.java;resolution:=optional,\
             io.vertx.rxjava.core.*;resolution:=optional,\
             io.vertx.rxjava.ext.auth.*;resolution:=optional,\
+            io.vertx.reactivex;resolution:=optional,\
```
cescoffier: Cant' we merge both instructions into one:

io.vertx.reactivex.*;resolution:=optional.

Is there other required packages that must be mandatory ?

pom.xml
sauthieg: I forgot them :)
thanks for the head-up

pom.xml
sauthieg: Well, there are MQTT classes generated in `io.vertx.reactivex.mqtt`, that seems weird to optionally import something that is already inside of the bundle

cescoffier: Is there a dependency on auth ? Some module relies on vertx-auth to handle the authentication.

ppatierno:
@cescoffier I'm not an OSGi expert so ... can I merge this PR after your change request now ? :)

cescoffier:
@ppatierno +1, go ahead. And thanks @sauthieg

pom.xml
sauthieg: No code in this bundle use auth, It should never have been there I think.
My bad, in my initial contribution, I forgot to remove them ..

PR #54, opened 2017-09-01
Implements MqttClientOptions(JsonObject) constructor

sauthieg:
I've seen that the `MqttClientOptions` constructors were empty.
So, here is a PR to fix that, avoiding surprises for end user :)


src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -74,7 +74,7 @@ public MqttClientOptions() {
    */
   public MqttClientOptions(JsonObject json) {
     super(json);
-    //TODO implement this constructor
+    MqttClientOptionsConverter.fromJson(json, this);
```
Sammers21: What is _MqttClientOptionsConverter_ ? Where is the class located?

sauthieg: It's been automatically generated with the help of the `@DataObject(generateConverter=true)` annotation.

As it's generated as part of the build, I've not that class under revision control.
If you're interested, generated source is being placed in `src/generated`.

Sammers21: Ok, great. Thank you for details.

pom.xml
```diff
@@ -106,10 +110,6 @@
       <artifactId>vertx-rx-java2</artifactId>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>io.vertx</groupId>
-      <artifactId>vertx-core</artifactId>
-    </dependency>
```
ppatierno: Why did you move this dependency from here moving it up ?

src/main/java/io/vertx/mqtt/MqttClientOptions.java
ppatierno: @sauthieg you can push this file in the PR don't worry.

pom.xml
sauthieg: because `vertx-codegen` includes `JsonObject` as well and it's missing the `getValue()` method. That was causing compilation error in my IDE.

It seems that codegen was first in the compilation classpath in my case (not sure if it's always the case)

src/main/java/io/vertx/mqtt/MqttClientOptions.java
sauthieg: Well, as it's being generated, I would prefer to keep it outside of rev control, otherwise for any new configuration attribute, we'll have to remember to copy the (changed) generated file in the sources.

ppatierno: Understood. @vietj what do you think about that ? For example in the Vert.x Kafka client we already have a generated file pushed to the repo (https://github.com/vert-x3/vertx-kafka-client/blob/master/src/main/generated/io/vertx/kafka/client/common/PartitionInfoConverter.java)

vietj: I think it's a misconfiguration issue in the project, it should be committed in the repo.

sauthieg: Ok, but we keep it in the `src/main/generated` directory ?
At least, git will show delta if something change in this file

ppatierno: Yes, we already have it in the Kakfa client repo.

vietj: it is actually wished to be committed as the source code depends on it, it allows someone that clones the project to find the source code already there.

sauthieg: updated

pom.xml
sauthieg: do you want me to keep this commit or not ?

Basically, the good solution would be to make sure that core classes embed in `vertx-codegen` are "compatible" with the real ones.

ppatierno: Comparing even with Kafka client poi.xml, the order works well there. It's weird that you needed to invert the order here ... or there is something I'm missing :)

sauthieg: :)
The error is only showing up in my IDE (IntelliJ). Maven compilation is not impacted and is successful, whatever ordering is used...
I'm perfectly happy to remove that commit if you wish.

ppatierno: Yes please ;)
Then I'll try with my IntelliJ as well.

sauthieg: done

Issue #53, opened 2017-08-31
Develop the persistence and recovery for "in flight" messages queues


ppatierno:
The client should be able to persist the "in flight" messages queues (QoS 1 and 2) in order to recover them after coming back online (due to a previous disconnection).

Sammers21:
I can work on that. But before should discuss it.

Let's start with a simple case:

1. The client was connected with a server and he just has some messages in queues. 
2. And then client was instantly disconnected from the server

And now there are some question to discuss: _What should be done next?_

I would propose that a user should:

1. catch closing a connection through a _closeHandler_
2. call _setCleanSession(true)_ 
3. call _connect_  

And after that, the client should one-by-one resend all packets in all queues

What do you think about it? Is that what is expected from the feature?

ppatierno:
Great ! :-)

You can check here (http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.pdf) how the client session works and it's based on the "clean session" flag sent within the CONNECT packet.

Sammers21:
@ppatierno, what do you think about the #57 PR?

PR #57, opened 2017-09-09
The persistence and recovery for "in flight" messages queues

Sammers21:
The PR should be a solution for #53


Sammers21:
@ppatierno, I have just added a test for QoS = 1 recovery. Is it ok? I would like to create a similar one for QoS =2

PR #58, opened 2017-09-09
Wait for server starting before running tests in MqttClientOutOfOrderAcksTest class

Sammers21:
For now, on my computer, such error occur when running tests from master: https://gist.github.com/Sammers21/55126fda925a58c879fed7a939231fce#file-log-txt-L704

As you seen in log, a client failed in connecting with server because in some cases it can be not listening related port while test running

### DIFF
--- a/src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
+++ b/src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
@@ -35,6 +35,7 @@ import org.junit.runner.RunWith;
 
 import java.util.LinkedList;
 import java.util.Queue;
+import java.util.concurrent.CountDownLatch;
 
 
 /**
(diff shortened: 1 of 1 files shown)


src/test/java/io/vertx/mqtt/test/client/MqttClientOutOfOrderAcksTest.java
```diff
@@ -93,18 +95,22 @@ private void clientSendThreePublishMessages(MqttQoS mqttQoS, TestContext context
   }
 
   @Before
-  public void before() {
+  public void before() throws InterruptedException {
+    vertx = Vertx.vertx();
```
ppatierno: I see that the vertx instance is already created on the declaration line : 

`Vertx vertx = Vertx.vertx();`

Sammers21: You are right

Sammers21:
squashed

PR #68, opened 2017-11-14
Improve the logging when connections fail

ctron:
This changed does lower the log output when a connection fails to WARN
as the application can always handle this properly. Also is the failed
state forwarded to the application specific part, which may decide if
this is an actual error.

Also the error code get added to the log message.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -825,7 +825,7 @@ public class MqttClientImpl implements MqttClient {
           this.connectHandler.handle(Future.succeededFuture(msg));
         } else {
           MqttConnectionException exception = new MqttConnectionException(msg.code());
-          log.error("Connection refused by the server");
+          log.error(String.format("Connection refused by the server - code: %s", msg.code()));
           this.connectHandler.handle(Future.failedFuture(exception));
         }
       }


src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat Inc.
+ * Copyright 2016, 2017 Red Hat Inc.
```
Sammers21: It is okay to leave 2016 for now. One day @ppatierno will update all headers

ppatierno: yes @Sammers21 is right, I have to update all the other files so revert this change please otherwise I'll have some mix :-)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -825,7 +825,7 @@ void handleConnack(MqttConnAckMessage msg) {
           this.connectHandler.handle(Future.succeededFuture(msg));
         } else {
           MqttConnectionException exception = new MqttConnectionException(msg.code());
-          log.error("Connection refused by the server");
+          log.warn(String.format("Connection refused by the server - code: %s", msg.code()));
```
ppatierno: imho it's an error because at MQTT application level your connection to the server didn't go well. It's something that the application should take care of because it's not connected to the server successfully. The warning in this case sounds to me something like "pay attention something went wrong but you can continue without any action". I would stick with error log level but updating the message as you are proposing with MQTT return code.

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
ctron: Copyright header should always be up to date 😉

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
ctron: Well the issue with that just is, your log might get spammed with error messages, although your application might be able to take care of this. So you need to manually suppress error messages for this MqttClientImpl.

But I did amend the PR and changed it back to "error". The error code is more important to me 😉

ctron:
Updated the PR according to the review.

Issue #66, opened 2017-11-14
MqttClient seems not to re-connect automatically


ctron:
The `MqttClientOptions` do have a field `reconnectAttempts`, but it seems as if this field is not actually used and the client does not automatically re-connect.

Sammers21:
Hi @ctron , actually automatic reconnection is not supported yet. But you can implement it with MqttClient#closeHandler. The field in `MqttClientOptions` is useless. 

If you'd like to fix it - feel free to contribute

ctron:
Well it can't be implemented 100% with the close handler as it looks to me.

When the connection succeeds, then the close handler will be called. However when the initial connection doesn't succeed, only the Future is completed with a failure:

~~~java
if (done.failed()) {
  log.error(String.format("Can't connect to %s:%d", host, port), done.cause());
  if (connectHandler != null) {
    connectHandler.handle(Future.failedFuture(done.cause()));
}
~~~

However, listening to the result of the future for a failure is not possible, because in this case you will also be noted over the closeHandler _and_ the Future. So for some (not for all) condition you might be notified twice for a broken connection.

I think this could be fixed by simply calling the close handler in addition to finishing the Future with a failure in the "if" mentioned above.

Sammers21:
That is a great idea and as a benefit, it would be way more easy to implement 100% reconnection. But, there is the other side... to call _closeHandler_ when we failed in the attempt to establish TCP connection. I think and it logical to not call _closeHandler_ when a connection has not been established yet so we have just nothing to close.

Sammers21:
Btw, @ppatierno, what do you think about it?

ctron:
Well if you don't call the closeHandler when it was not connected, then you can't implement re-connect that way. But if you always do call the closeHandler, then you can check locally (in your state) if you had a successful connection before.

The most important thing might be some proper documentation in the javadoc to explain this. If this is all that it needs, I am happy to amend the PR for more documentation.

vietj:
keep in mind that Vert.x NetClient already supports connect retries when the connection fails (not when it is disconnected as this is protocol related)

ctron:
But I did understand @Sanmers21 that this is not supported with the MQTT client.

ppatierno:
[...]

PR #69, opened 2017-11-14
#66: Fix missing callback when TCP connect fails

ctron:
When the TCP connect fails, then there is no notification via the
closeHandler. On the other side listening to the connect future for
errors notifies you twice of an error (via closeHandler and via future)
in cases where the TCP connect succeeded, but a CONACK with an error
is received.

This change adds a call to the closeHandler in case the TCP connect
fails, so consumers can solely rely on the closeHandler for failed
connection attempts and can re-trigger a connection, thus implementing
"auto reconnect".

Signed-off-by: Jens Reimann <jreimann@redhat.com>

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Red Hat Inc.
+ * Copyright 2016, 2017 Red Hat Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -165,6 +165,10 @@ public class MqttClientImpl implements MqttClient {
         log.error(String.format("Can't connect to %s:%d", host, port), done.cause());
         if (connectHandler != null) {
           connectHandler.handle(Future.failedFuture(done.cause()));
+        }
+        final Handler<Void> closeHandler = this.closeHandler;
+        if (closeHandler != null) {
+            closeHandler.handle(null);
         }
       } else {
         log.info(String.format("Connection with %s:%d established successfully", host, port));


ctron:
The connectHandler isn't enough when the connection breaks. So you have the following scenarios:

### TCP connect fails

* connectHandler gets fired with failed

### TCP connect succeeds, CONACK is failed

* connectHandler gets fired with failed
* closeHandler gets fired

### all succeeds, but fails later with TCP disconnect

* closeHandler gets fires

----

So now you are unable to handle failed -> reconnect in sane way. Because there is not clear path of notifying a connection failure. Especially in scenario 2, you would need to filter out one of the notifications. But you can't really do that, since you don't know if the other notification will get fired.

If however you would fire the close handler in scenario 1 as well, you could always handle connection failures via the path of closeHandler.

ctron:
Sure here is the code I am using, under the assumption this PR has been applied: https://github.com/ctron/flow/blob/develop/components/mqtt/src/main/java/de/dentrassi/flow/component/mqtt/MqttClient.java#L98

No I can't solely rely on the `connectHandler` as then I would not get notified when the connection breaks after the connection has been successfully established. The problem is with the scenario in the middle. When the TCP connection has been established, but the MQTT connection has not. In this case both handlers are fired.

ppatierno:
@ctron imho the `closeHandler` makes sense when a connection is established and is working and then it's closed by the server or goes down for any other network reason. It doesn't make more sense on TCP connection establishment failure. I know that it sounds strange but maybe the reconnect feature should be developed inside the `closeHandler`; in the case when the TCP connection fails it should happen in the `connectHandler`. I guess that in this way you can cover both the scenarios without any changes to the `MqttClient`, right ?

vietj:
@ppatierno for close handler, I meant as complement to the built-in connection retry failure that NetClient does OOTB. It can either be externally done or built in the MQTT client.

vietj:
i.e in `MqttClientImpl`

```
  void handleClosed() {
    synchronized (this.connection) {
      if (this.closeHandler != null) {
        this.closeHandler.handle(null);
      }
    }
  }
```

ctron:
Ok, we just had a call … The problem is the scenario "TCP connect succeeds, CONACK is failed" where you get two callbacks with the meaning of "connection failed". @ppatierno will think of something 😉

ctron:
So would that mean that in the case 2./c) only the connectHandler gets fired?

ctron:
Cool .. that would work fine!

ctron:
Closing this out in favor of #71

PR #71, opened 2017-11-22
Handling connections failure for allowing "manual" reconnection

ppatierno:
Added connection status in the MqttClient interface
Added check for not calling the closeHandler on negative CONNACK
Added some unit tests for handling connection failures

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -829,10 +829,11 @@ public class MqttClientImpl implements MqttClient {
 
     synchronized (this.connection) {
 
+      this.isConnected = msg.code() == MqttConnectReturnCode.CONNECTION_ACCEPTED;
+
       if (this.connectHandler != null) {
 
         if (msg.code() == MqttConnectReturnCode.CONNECTION_ACCEPTED) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -822,6 +832,7 @@ void handleConnack(MqttConnAckMessage msg) {
       if (this.connectHandler != null) {
 
         if (msg.code() == MqttConnectReturnCode.CONNECTION_ACCEPTED) {
+          this.isConnected = true;
```
ctron: As the user might choose to provide `null` as `connectHandler`doing the assignment inside the `if (this.connectHandler != null)` would result in `isConnected()` to return `false` in such cases.

I think the assignment should be re-written in a way that it takes care of this, like:

~~~java
this.isConnected = msg.code() = MqttConnectReturnCode.CONNECTION_ACCEPTED;

if (this.connectHandler != null) {
…
~~~

ppatierno: You are absolutely right !

ppatierno: You are absolutely right !

Issue #78, opened 2018-01-18
Using `setIdleTimeout` causes duplicate filter exception


ctron:
Setting the idle timeout for a client using:

~~~java
final MqttClientOptions options = new MqttClientOptions();
options.setIdleTimeout(10);
~~~

Causes the following exception:

~~~
14:31:09.663 [vert.x-eventloop-thread-1] ERROR io.vertx.core.impl.ContextImpl - Unhandled exception
java.lang.IllegalArgumentException: Duplicate handler name: idle
	at io.netty.channel.DefaultChannelPipeline.checkDuplicateName(DefaultChannelPipeline.java:1071)
	at io.netty.channel.DefaultChannelPipeline.filterName(DefaultChannelPipeline.java:302)
	at io.netty.channel.DefaultChannelPipeline.addBefore(DefaultChannelPipeline.java:259)
	at io.netty.channel.DefaultChannelPipeline.addBefore(DefaultChannelPipeline.java:249)
	at io.vertx.mqtt.impl.MqttClientImpl.initChannel(MqttClientImpl.java:608)
	at io.vertx.mqtt.impl.MqttClientImpl.lambda$doConnect$2(MqttClientImpl.java:180)
	at io.vertx.core.net.impl.NetClientImpl.lambda$connect$2(NetClientImpl.java:118)
	at io.vertx.core.net.impl.NetClientImpl.lambda$null$9(NetClientImpl.java:256)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:342)
	at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:200)
	at io.vertx.core.net.impl.NetClientImpl.lambda$connected$10(NetClientImpl.java:252)
	at io.vertx.core.net.impl.VertxHandler.setConnection(VertxHandler.java:50)
	at io.vertx.core.net.impl.VertxNetHandler.handlerAdded(VertxNetHandler.java:44)
	at io.netty.channel.DefaultChannelPipeline.callHandlerAdded0(DefaultChannelPipeline.java:606)
	at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:235)
	at io.netty.channel.DefaultChannelPipeline.addLast(DefaultChannelPipeline.java:201)
	at io.vertx.core.net.impl.NetClientImpl.connected(NetClientImpl.java:262)
[...]

PR #79, opened 2018-01-18
deprecate setIdleTimeout and force it to do the same as setKeepAliveTimeSeconds

Sammers21:
A solution for #78

### DIFF
--- a/src/main/java/io/vertx/mqtt/MqttClientOptions.java
+++ b/src/main/java/io/vertx/mqtt/MqttClientOptions.java
@@ -382,6 +382,15 @@ public class MqttClientOptions extends NetClientOptions {
     return this;
   }
 
+  /**
+   * Do the same thing as {@link MqttClientOptions#setKeepAliveTimeSeconds(int)}. Use it instead.
+   */
+  @Deprecated
+  @Override
+  public MqttClientOptions setIdleTimeout(int idleTimeout) {
+    return setKeepAliveTimeSeconds(idleTimeout);
+  }
+
   @Override
   public MqttClientOptions setSsl(boolean ssl) {
     super.setSsl(ssl);
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -47,6 +47,7 @@ import io.vertx.core.impl.NetSocketInternal;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import io.vertx.core.net.NetClient;
+import io.vertx.core.net.NetClientOptions;
 import io.vertx.mqtt.MqttClient;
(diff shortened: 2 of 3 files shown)


Sammers21:
@ppatierno, just done.

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -382,6 +382,15 @@ public MqttClientOptions setMaxMessageSize(int maxMessageSize) {
     return this;
   }
 
+  /**
+   * Do the same thing as {@link MqttClientOptions#setKeepAliveTimeSeconds(int)}. Use it instead.
+   */
+  @Deprecated
+  @Override
+  public NetClientOptions setIdleTimeout(int idleTimeout) {
```
ppatierno: the return type has to be `MqttClientOptions`

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -82,6 +83,7 @@
   private static final int MIN_TOPIC_LEN = 1;
   private static final String PROTOCOL_NAME = "MQTT";
   private static final int PROTOCOL_VERSION = 4;
+  private static final int DEFAULT_INTERNAL_NET_CLIENT_IDLE_TIMEOUT = 0;
```
ppatierno: I think that `DEFAULT_IDLE_TIMEOUT` is enough

src/test/java/io/vertx/mqtt/test/client/MqttClientConnectTest.java
```diff
@@ -59,9 +59,11 @@ public void connectDisconnect(TestContext context) throws InterruptedException {
   }
 
   @Test
-  public void connectDisconnectNoOptions(TestContext context) {
+  public void connectDisconnectWithIdleOption(TestContext context) {
```
ppatierno: Why did you change the `connectDisconnectNoOptions` already in place ? Please create a new test for this fix.

Sammers21: The funny thing is that _connectDisconnectNoOptions_ is a full duplicate of _connectDisconnect_ so i have decided to modify it instead of removing as duplicate :)

ppatierno: ah !! You are right :-) So it's ok ... see only the other request changes I did

src/main/java/io/vertx/mqtt/MqttClientOptions.java
Sammers21: Sure

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
Sammers21: Agree

Sammers21:
squashed

Issue #81, opened 2018-02-11
paho client in python  won't work


Ashye:
when i connect my paho client (in python) to vertx mqtt server, it always raise the same error:
```
java.lang.Exception: Wrong message type
	at io.vertx.mqtt.impl.MqttServerConnection.handleMessage(MqttServerConnection.java:178)
	at io.vertx.mqtt.impl.MqttServerImpl.lambda$null$3(MqttServerImpl.java:89)
	at io.vertx.core.net.impl.NetSocketImpl.handleMessageReceived(NetSocketImpl.java:356)
	at io.vertx.core.net.impl.NetServerImpl$2.handleMessage(NetServerImpl.java:445)
	at io.vertx.core.net.impl.NetServerImpl$2.handleMessage(NetServerImpl.java:442)
	at io.vertx.core.net.impl.VertxHandler.lambda$channelRead$1(VertxHandler.java:150)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:342)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:80)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

Sammers21:
@Ashye, can you provide a reproducer and upload it on Github?

Ashye:
i found out the pattern of the issue.
when i deploy a mqtt server with deploy option go with setWorker(true), the problem appearred.
but i don't know why?

Sammers21:
Looking at the error, it seems that message received by mqtt-server is not related to mqtt... 

It is also hard to figure out where is the problem without having a reproducer.

I also don't think that `setWorker(true)` may affect, but who knows.

Ashye:
this is my testing code.
clean project after changing setworker(), 
 

```
class DemoServerVerticle : BaseVerticle(true) {

    companion object {
        @JvmStatic
        fun main(args: Array<String>) {
            Vertx.factory.vertx().deployVerticle(DemoServerVerticle::class.java.canonicalName,
                    DeploymentOptions().setWorker(true))
        }
    }


    private val mqttServer: MqttServer by lazy { MqttServer.create(vertx) }
    /**
     * client subscription topic list
     */
[...]

PR #84, opened 2018-02-22
Print class name when type of message is wrong

Sammers21:
Might also help to discover cause of #81

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -172,14 +172,14 @@ public class MqttServerConnection {
 
         default:
 
-          this.chctx.fireExceptionCaught(new Exception("Wrong message type " + msg.getClass().getName()));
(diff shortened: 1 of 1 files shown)


src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -179,7 +179,7 @@ synchronized void handleMessage(Object msg) {
 
     } else {
 
-      this.chctx.fireExceptionCaught(new Exception("Wrong message type"));
+      this.chctx.fireExceptionCaught(new Exception("Wrong message type " + msg.getClass().getName()));
```
ppatierno: Yes I think that it could be useful here.
I also noticed that in the `default` case above there is the same message which is not useful because we know that the class is an `MqttMessage` and we should print the message type using `mqttMessage.fixedHeader().messageType()`. Can you make this change and then squashing this two commits ? I'll be happy to merge :-)

Sammers21: Done

PR #86, opened 2018-03-02
MQTT server publish overload

Sammers21:
a solution for #85

PR #87, opened 2018-03-03
fixed #85  require an messageId and send message with it

### LINKED ISSUE #85: how I kown the server publish message （use qos1） is success
when I publish message1 use method
`  MqttEndpoint publish(topic, Buffer payload, qosLevel, isDup, isRetain);  `

and i can get  messageId by method
  `int lastMessageId()`

and i kown one message is success when i get
 ```
client.publishAcknowledgeHandler(id->{
    String msg = getMailBox(client).getMsg(id);
    System.out.println("success:"+id+", msg:"+msg);
});
```

but in multi thread , the messageId maybe not the real messageId, when an other thread send message2 at same time

so if i want get real messageId for message1, i must use lock,  is that?

### PR DESCRIPTION
**Pull request by caoqingguang**
fixed #85  require an messageId and send message with it, when get ack from remote client , will kown the message is success.  if no ack for a long time , i can try again send the same message with the same messageId .

### DIFF
--- a/src/main/java/io/vertx/mqtt/MqttEndpoint.java
+++ b/src/main/java/io/vertx/mqtt/MqttEndpoint.java
@@ -353,6 +353,24 @@ public interface MqttEndpoint {
   MqttEndpoint publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain);
 
   /**
+   * Sends the PUBLISH message with specified messageId to the remote MQTT client
+   *
+   * @param topic    topic on which the message is published
+   * @param payload  message payload
+   * @param qosLevel quality of service level
+   * @param isDup    if the message is a duplicate
+   * @param isRetain if the message needs to be retained
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  MqttEndpoint publishWithId(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain,int messageId);
+
+  /**
+   * require an messageId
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/main/java/io/vertx/mqtt/MqttEndpoint.java [hunk 1]
```diff
@@ -352,6 +352,24 @@
...
+   *
+   * @param topic    topic on which the message is published
+   * @param payload  message payload
+   * @param qosLevel quality of service level
+   * @param isDup    if the message is a duplicate
+   * @param isRetain if the message needs to be retained
+   * @return a reference to this, so the API can be used fluently
+   */
+  @Fluent
+  MqttEndpoint publishWithId(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain,int messageId);
```
Sammers21: How can a message Id generated by hands be useful?

src/main/java/io/vertx/mqtt/MqttEndpoint.java
```diff
@@ -352,6 +352,24 @@
...
+
+  /**
+   * require an messageId
+   * @return the next messageId
+   */
+  int requireMessageId();
```
Sammers21: What is a purpose of this?

ppatierno:
I do think that giving the possibility to the user to set the message-id can lead to a lot of problems even considering that I should review the @Sammers21 PR (I know Pavel ... you hate me ;)) about in-flight messages. The message-id is for internal purpose of the protocol I think that PR #86 makes more sense.

PR #89, opened 2018-04-12
GitHub issue template

Sammers21:
When creating an issue it is a very common case when reporter provides lack of information on how to reproduce the actual problem. The proposed template should address the issue.


Sammers21:
@vietj , @ppatierno , can you review?

Issue #92, opened 2018-05-14
Deadlock in io.vertx.mqtt.impl.MqttClientConnection


kirillkh:
Vertx 3.5.1

After several days of running my server that uses vertx-mqtt, it got stuck in a deadlock with the following stack trace. The verticle that runs this client is really simple, it only calls `client.publish` in a single place, nothing else. I'm fairly sure there is no possibility my code is to blame for this deadlock.

Looking at my logs, I found that the other thread that apparently participated in this deadlock was not managed by Vertx, so there is no stack trace available, but I can tell it got locked either when calling `client.subscribe(String, int, Handler)` or waiting the subscription to complete on a *different* instance of MqttClient running in a different verticle.  I was not able to extract the other stack trace, because I was in a rush to restart the server.

Any help would be appreciated.

```
[2018-05-14 11:06:20] [WARNING] io.vertx.core.impl.BlockedThreadChecker: Thread Thread[vert.x-eventloop-thread-2,5,main] has been blocked for 5246 ms, time limit is 2000
io.vertx.core.VertxException: Thread blocked
        at io.vertx.mqtt.impl.MqttClientConnection.handleMessage(MqttClientConnection.java:61)
        at io.vertx.mqtt.impl.MqttClientImpl.lambda$null$0(MqttClientImpl.java:192)
        at io.vertx.mqtt.impl.MqttClientImpl$$Lambda$64/57249380.handle(Unknown Source)
        at io.vertx.core.net.impl.NetSocketImpl.handleMessageReceived(NetSocketImpl.java:351)
        at io.vertx.core.net.impl.NetClientImpl$1.handleMessage(NetClientImpl.java:242)
        at io.vertx.core.net.impl.NetClientImpl$1.handleMessage(NetClientImpl.java:239)
        at io.vertx.core.net.impl.VertxHandler.lambda$channelRead$1(VertxHandler.java:146)
        at io.vertx.core.net.impl.VertxHandler$$Lambda$71/465081768.run(Unknown Source)
        at io.vertx.core.impl.ContextImpl.lambda$wrapTask$2(ContextImpl.java:337)
        at io.vertx.core.impl.ContextImpl$$Lambda$29/199640888.run(Unknown Source)
        at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:195)
        at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:144)
[...]

Issue #82, opened 2018-02-11
mqtt2kafka application run a deadlock for a period of time


lifengchuan2008:
- vertx  3.5.0
- netty  4.1.8.Final

I have a problem，the mqtt to Kafka application receives 51 messages every 2 seconds after 20 hours of deadlock,

#### MqttService
```
  Future<MqttClient> mqttClientFuture = Future.future();
        vertx.executeBlocking(future -> {
            client = MqttClient.create(vertx, options);
            future.complete(client);
        }, res -> {
            mqttClientFuture.complete((MqttClient) res.result());
        });
        mqttClientFuture.setHandler(result -> {
            client.rxConnect(port, host).flatMap(ack -> client.rxSubscribe(topic, 0)).subscribe(sr -> {
                log.debug("mqtt  connected  ,packetId: {}", sr);
            }, error -> {
                log.error("mqtt  connect error: {}", error);
            });
            log.debug("client: {},context: {}", client, vertx.getDelegate().getOrCreateContext());
            client.publishHandler(message -> {
              
                    kafkaService.sendMsg(kafkaTopic, message.payload().toString().getBytes());
                
            });
```

#### KafkaService

```
    @Override
    public void sendMsg(String topic, byte[] msg) {
        producer.rxWrite(KafkaProducerRecord.create(topic, msg)).subscribe(metadata -> {
        });
    }
```

vert.x-worker-thread-4      Mqtt subscribes to tiopic and accepts mqtt messages

 vert.x-eventloop-thread-1    The message received by mqtt is forwarded to Kafka

```
Found one Java-level deadlock:
=============================
"vert.x-worker-thread-4":
  waiting to lock monitor 0x00007efef4020788 (object 0x0000000086964490, a io.vertx.mqtt.impl.MqttClientConnection),
  which is held by "vert.x-eventloop-thread-1"
"vert.x-eventloop-thread-1":
  waiting to lock monitor 0x00007efefc0287d8 (object 0x00000000869644b0, a io.vertx.core.net.impl.NetSocketImpl),
  which is held by "vert.x-worker-thread-4"

Java stack information for the threads listed above:
===================================================
"vert.x-worker-thread-4":
	at io.vertx.mqtt.impl.MqttClientConnection.handleMessage(MqttClientConnection.java:61)
	- waiting to lock <0x0000000086964490> (a io.vertx.mqtt.impl.MqttClientConnection)
[...]

PR #96, opened 2018-05-25
Use NetSocket as lock on MqttClientConnection instead of itself

### LINKED ISSUE #93: Use NetSocket as lock on MqttClientConnection instead of itself
Which would avoid deadlocks.

Sammers21:
fixes #93 #92 #82


Sammers21:
@vietj, can you review?

vietj:
can you check, since we merged the SSL PR, there is a new synchronized usage for the SSL getter/setter

Sammers21:
Yes, I have just checked. There is a new synchronization related to the MQTT-server, not for the client component, so it is fine.

PR #101, opened 2018-06-11
fix MqttServer Duplicate handler name: idle

dushibaiyu:
When  The NetServerOptions set the ' IdleTimeout'. The connection will add the name 'idle' handler before.
And the mqttserver will add the same name handler too.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
@@ -138,7 +138,7 @@ public class MqttServerImpl implements MqttServer {
     }
 
     // adding the idle state handler for timeout on CONNECT packet
-    pipeline.addBefore("handler", "idle", new IdleStateHandler(this.options.timeoutOnConnect(), 0, 0));
+    pipeline.addBefore("handler", "mqttIdle", new IdleStateHandler(this.options.timeoutOnConnect(), 0, 0));
     pipeline.addBefore("handler", "timeoutOnConnect", new ChannelDuplexHandler() {
 
       @Override


Sammers21:
@dushibaiyu , can you implement in a way, as it done in the MQTT client, I mean deprecating `MqttServerOptions#setIdleTimeout` and delegating it to the `MqttServerOptions#setTimeoutOnConnect` call.

dushibaiyu:
Yes。 it changed.

src/main/java/io/vertx/mqtt/MqttServerOptions.java
```diff
@@ -252,10 +252,29 @@ public MqttServerOptions setTimeoutOnConnect(int timeoutOnConnect) {
...
   public int timeoutOnConnect() {
     return this.timeoutOnConnect;
   }
+
+  /**
+   * @return  the TCP idle timeout, in seconds,In MqttServer the Parameter should be 0, it will use timeoutOnConnect.
+   */
+  @Override
+  public int getIdleTimeout() {
+    return 0;
```
Sammers21: Why returning 0 here? The method will return zero by default.

And, thinking about it... I think we can also delegate the method call to `MqttServerOptions#timeoutOnConnect`

src/main/java/io/vertx/mqtt/MqttServerOptions.java
```diff
@@ -252,10 +252,29 @@ public MqttServerOptions setTimeoutOnConnect(int timeoutOnConnect) {
     return this;
   }
 
+  /**
+   * Do the same thing as {@link MqttClientOptions#setKeepAliveTimeSeconds(int)}. Use it instead.
+   */
+  @Deprecated
+  @Override
+  public MqttServerOptions setIdleTimeout(int idleTimeout) {
+    super.setIdleTimeout(0);
```
Sammers21: Why calling the super method here? I think we should not do this.

src/main/java/io/vertx/mqtt/MqttServerOptions.java
```diff
@@ -252,10 +252,29 @@ public MqttServerOptions setTimeoutOnConnect(int timeoutOnConnect) {
     return this;
   }
 
+  /**
+   * Do the same thing as {@link MqttClientOptions#setKeepAliveTimeSeconds(int)}. Use it instead.
```
Sammers21: Do the same thing as `{@link MqttClientOptions#setTimeoutOnConnect(int)}`. Use it instead.

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -154,5 +154,6 @@ public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exc
         }
       }
     });
+
```
Sammers21: Can you please remove this line? Since it is not related to the PR

dushibaiyu:
Yes , I have fixed it.

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
dushibaiyu: ok

src/main/java/io/vertx/mqtt/MqttServerOptions.java
```diff
@@ -252,10 +252,21 @@ public MqttServerOptions setTimeoutOnConnect(int timeoutOnConnect) {
...
+  }
+
+
   /**
    * @return  timeout on CONNECT before closing connection
    */
   public int timeoutOnConnect() {
     return this.timeoutOnConnect;
   }
+
```
Sammers21: Can you remove the line?

src/main/java/io/vertx/mqtt/MqttServerOptions.java
```diff
@@ -252,10 +252,21 @@ public MqttServerOptions setTimeoutOnConnect(int timeoutOnConnect) {
...
+  /**
+   * Do the same thing as {@link MqttServerOptions#setTimeoutOnConnect(int)}. Use it instead.
+   */
+  @Deprecated
+  @Override
+  public MqttServerOptions setIdleTimeout(int idleTimeout) {
+    return setTimeoutOnConnect(idleTimeout);
+  }
+
+
```
Sammers21: And please remove one empty line here.

dushibaiyu: ok

src/main/java/io/vertx/mqtt/MqttServerOptions.java
Sammers21: @dushibaiyu , don't forget to remove the line, please.

Issue #109, opened 2018-11-13
mqtt client will never realize the connection is closed when TCP connection established but CONACK not received


SunFeiyue:
For the following situation :
1) the underlying TCP connection is established
2) client receives no CONACK
3) the underlying TCP connection is disconnected

In this situation, the mqtt client "connect handler" will not be called because CONACK is not received, the mqtt client 'close handler' will not be called because 'isConnected' is false. So the client will never realize the connect result because it doesn't receive any feedback, except monitoring the 'isConnected' value.

vietj:
why does it matter since that's not a real connection ?

SunFeiyue:
> 
> 
> why does it matter since that's not a real connection ?
Since the client dosen't receive any feedback, the client code will not be triggered. 
We have met such a problem: we have implemented auto-connect function when the mqtt connection is broken. We wrote the 'connect-retry' logic in the 'close handler' and 'connect handler'. In 'connect handler', we check that if the connect fails, we will do 'connect-retry' after a period of time. In 'close handler', we will also do 'connect-retry' after a period of time. In the above situation, since the connect handler and close handler will not be called, our 'auto-connect' mechanism will fail.

ppatierno:
@SunFeiyue I think that it makes sense. The close handler is supposed to be called when the TCP connection is closed and should not be tied to the CONNACK. Thanks for raising it.
Are you going to provide a PR for this one?

vietj:
@ppatierno ok, so that would mean that the client sends at least a CONNECT message right ?

vietj:
I thought the client would not send anything first :-)

ppatierno:
@vietj the client should do that: opening the TCP connection, sending the CONNECT and waiting for the CONNACK. I think that the problem they are facing is about the CONNECT cannot reach the server (so it doesn't send CONNACK) or the client doesn't receive a CONNACK and somehow the server closes the TCP connection (or it's interrupted). Is that right @SunFeiyue ?

vietj:
if CONNECT does not reach the server then it will not help because the handler won't get MqttEndpoint and won't have opportunity to set a close handler

ppatierno:
Yeah agree but @SunFeiyue is talking about "the underlying TCP connection is disconnected" so it should be detected at Netty "level" on the client and the close handler should be called. Of course, if the TCP connection is not disconnected, the client could wait indefinitely because it can be related to the CONNECT or CONNACK lost but with the TCP connection still in place.

[...]

PR #112, opened 2018-11-14
Mqtt client read timeout

Sammers21:
A solution for #109.
Duplicate of #110.


Sammers21:
@ppatierno , @vietj , can you please review?

Issue #117, opened 2019-01-15
JAVA: keep alive not working


konradmichael:
The keep alive does not seem to work, maybe i also found out why it is so.

Reproduce:
1. Connect to any remote mqtt-broker (not on your pc).
2. unplug network cable
Result:
Nothing. The connection will not be closed

If you replug the network, the connection will be closed and reconnected (in my setup).

Potential error: MqttClientImpl line 140
https://github.com/vert-x3/vertx-mqtt/blob/c8213e0f41150a1a02f4ad1185762494b52c6350/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java#L140
`public MqttClientImpl(Vertx vertx, MqttClientOptions options) {

    // copy given options
    NetClientOptions netClientOptions = new NetClientOptions(options);
    netClientOptions.setIdleTimeout(DEFAULT_IDLE_TIMEOUT);

    this.client = vertx.createNetClient(netClientOptions);
    this.options = options;
}`

The `DEFAULT_IDLE_TIMEOUT` is **0**. Zero disabled timeout detection, I belive this should be `keepAlive * 1.5`

My current workaround is:

`
Future<Void> disconnectTask;
ScheduledExecutorService scheduledExecutorService;
int keepAlive = 30;
int timeout = 45;

// keep alive with multiplier to prevent multiple pings
mqttClientOptions.setKeepAliveTimeSeconds(keepAlive * 2);

mqttClient.pingResponseHandler(e ->  {
  // cancel timeout task
	disconnectTask.cancel(false);
  // schedule new timeout task
	disconnectTask = scheduledExecutorService.schedule(() -> {
		logger.warn("disconnecting because keep alive fails!");
		client.disconnect();
		return null;
	}, timeout, TimeUnit.SECONDS);
	// keep alive will not be send if other messages are processed,
	// force sending or it will get complicated!
	scheduledExecutorService.schedule(() -> client.ping(), keepAlive, TimeUnit.SECONDS);
}
`

I would be glad to receive some feedback

konradmichael:
Pull request for working solution: https://github.com/vert-x3/vertx-mqtt/pull/118

vietj:
can you elaborate what does the 1.5 factor does ? it is not clear to me what is the actual issue and what the fix changes.

konradmichael:
Issue: Client will not detect a missing ping response as a disconnect.
Fix: terminates connection if no response is comming from broker (timeout)

[...]

PR #118, opened 2019-01-16
close connection if keep-alive is exceeded by factor 1.5

konradmichael:
Solution for #117


vietj:
the 1.5 factor applies to the server and not to the client according to the spec.

Issue #123, opened 2019-03-20
Auto keep alive not working


ctron:
The auto keep alive seems not to be working. Checking the source I assume this is due to the following way it is implemented:

~~~java
pipeline.addBefore("handler", "idle",new IdleStateHandler(0, this.options.getKeepAliveTimeSeconds(), 0));
~~~

This will register an idle detector which only listens for write events. However, opening a connection and only pushing QoS messages will not trigger a response from the server,
so the client does never receive anything from the server.

The client should send a PINGREQ when the **reader** gets idle. Which shouldn't happen
when you send QoS 1 or 2 messages and the server sends responses. However if it does,
then sending a PINGREQ would be the appropriate action.

If the server doesn't response then, it should be considered dead.

vietj:
@ctron so this is mainly an mqtt-client concern right ?

ctron:
> @ctron so this is mainly an mqtt-client concern right ?

Yes, this is an issue in the MQTT client.

vietj:
so basically we should set a timer on the client to periodically send PING commands to the server and that's it right ?

ctron:
> so basically we should set a timer on the client to periodically send PING commands to the server and that's it right ?

Well you don't need a timer, a simple "reader idle" condition should work, swapping that with the current "writer idle".

vietj:
what do you mean ?

ctron:
Swapping the current reader idle condition with writer idle …

vietj:
ahhh... that makes sense lol :-) sorry

vietj:
how can we best make a reproducer test for this ?

ctron:
> how can we best make a reproducer test for this ?

I am not that familiar with testing in vertx. However the test should be:

Common start

* Create a connection, with an auto-handshake period of X seconds
* Start sending QoS 0 messages to the server, every 1 second
* After X seconds, the client should automatically send a PINGREQ

Variant A:

* The server does not respond
* The connection must be closed after X mutliplied by 1.5 seconds (MQTT spec)

Variant B:

* The server responds with a PINGRESP
* After X more seconds, the connection must still be open

vietj:
that's what I actually asked for :-) thanks.

ctron:
> that's what I actually asked for :-) thanks.

Cool :smile: … thanks for looking into this!

vietj:
[...]

PR #129, opened 2019-03-28
integrated client side keep alive timeout

konradmichael:
This is my implementation for fixing the Issue #123. This time it also includes Unit-Tests

There are three main changes:
1. Introduction of parameter 'keepAliveTimeout' (default:10) - Code comes from 'vietj's branch
2. Added IdleStateHandler for closing the connection, if no input is coming within timeout
3. keepAliveHandler will react on READER_IDLE

to Point3:
If you only send messages with QoS-0 you wont get anything from the broker. This way you cannot identify the state of the server connection. If you need to know if the server is responding you have to include the READER_IDLE.
Consequence of this is, that you usually send two PINGREQ, one for WRITER_IDLE and one for READER_IDLE.

I appreciate any help on this topic


vietj:
keepAliveTimeout is for the server not the client, also all tests have not been ported

vietj:
sorry you just renamed on the server the 1.5 factor variable to `keepAliveTimeout` which was not clear initially

vietj:
I also did a modification of the MqttServerImpl to support disabling of responses to test the case where the server will not send a ping response in time, can you port this + the test as well to have this case covered ?

konradmichael:
> keepAliveTimeout is for the server not the client, also all tests have not been ported
I have no Idea what that means? Can you please be more clear? What tests have to be ported?

> sorry you just renamed on the server the 1.5 factor variable to `keepAliveTimeout` which was not clear initially
Sometimes it is more helpful to use a meaningful variable name, `timeout` is very general. But if you want it to be reverted, just say so
It is a leftover of some test implementation.

> I also did a modification of the MqttServerImpl to support disabling of responses to test the case where the server will not send a ping response in time, can you port this + the test as well to have this case covered ?

Yes you made a modification. But after reviewing it, I have not found the use of it, that's why I did not integrate it. All the option did, was to stop the server-side timeout. It did not disable all keepAlive handling (PINGREQ / PINGRESP). So the option name was also misleading.
Also you see, that the test still worked without that option, so again why would you need that option.

I also tried to make some more tests, but they were full of magic numbers and the outcome would not have been meaningful. E.g. How do you test that your connection does NOT timeout ? :)
Testing is an art ...

konradmichael:
The travis checks are failing because of a blocked port :)
Problem here seems to be the parallel run of multiple builds / tests. It might be a good idea to revert the tests to use dynamic free ports.

```java
    try (ServerSocket socket = new ServerSocket(0)) {
        socket.setReuseAddress(true);
        return socket.getLocalPort();
    }
```

vietj:
I see what you mean now, indeed it is not necessary, in my test I made the mistake to use a `setKeepAliveTimeout` larger than `setKeepAliveTimeSeconds` on the client which required this change.

I'll make a few comments in the PR.

You want to test that the client does not timeout because it receives ping in time from the server ?

src/test/java/io/vertx/mqtt/test/server/MqttServerKeepAliveTest.java
```diff
@@ -0,0 +1,80 @@
...
+    MqttClientOptions options = new MqttClientOptions();
+    options.setAutoKeepAlive(false);    // The client will manage pings manually
+    options.setKeepAliveTimeSeconds(2); // Tell the server to disconnects the client after 3 seconds of inactivity
+    MqttClient client = MqttClient.create(vertx, options);
+    client.connect(1884, "localhost", ctx.asyncAssertSuccess(ack -> {
+      client.closeHandler(v -> {
+        async.complete();
+      });
+    }));
+    async.await(4000);
```
vietj: this is not needed

src/test/java/io/vertx/mqtt/test/client/MqttClientKeepAliveTest.java
```diff
@@ -0,0 +1,88 @@
...
+    options.setKeepAliveTimeSeconds(1);
+    options.setKeepAliveTimeout(1);
+    MqttClient client = MqttClient.create(vertx, options);
+    client.connect(1884, "localhost", ctx.asyncAssertSuccess(ack -> {
+      client.closeHandler(v -> {
+        assertEquals(2, pings.get()); // READER & WRITER idle
+        async.complete();
+      });
+    }));
+    async.await(4000);
```
vietj: this is not needed

src/test/java/io/vertx/mqtt/test/client/MqttClientKeepAliveTest.java
```diff
@@ -0,0 +1,88 @@
...
+    AtomicInteger pings = new AtomicInteger();
+    server.endpointHandler(endpoint -> {
+      endpoint.autoKeepAlive(false); // Tell the server not to respond to PINGREQ
+      endpoint.accept(false);
+      endpoint.pingHandler(v -> pings.incrementAndGet());
+    });
+    startServer(ctx);
+    Async async = ctx.async();
+    MqttClientOptions options = new MqttClientOptions();
+    options.setKeepAliveTimeSeconds(1);
```
vietj: use a larger value than 1

Currently this will tell the server to disconnect after 1 seconds it receives a ping, as the client sends a ping after 1 second, so the server might try to disconnect after 2 seconds, this can race with the fact that we test the client will close the connection and not the server, which can happen in CI and report false negatives.

I think the value `6` is a good trade off: the client will disconnect after 7 seconds (6 + 1) and the server would disconnect after 9 seconds which sounds which gives a 2 seconds delta.

src/test/java/io/vertx/mqtt/test/server/MqttServerKeepAliveTest.java
konradmichael: ok :) But I would like to keep the timeout.
The last time i was running this test and it failed, it did not stop. I havent tried to find the default timeout.
If you know a better way, feel free to change

src/test/java/io/vertx/mqtt/test/client/MqttClientKeepAliveTest.java
konradmichael: You are right. I usually try to keep these values as tiny as possible to reduce test runtime.

konradmichael:
> You want to test that the client does not timeout because it receives ping in time from the server ?

That is a complex test with many magic numbers. From my point of view it is very hard to make such a test that looks reasonable. What interval is long enough to wait, so that this is a real test ...

It is always pretty tough, to test that nothing happens :)

I will update the PR on monday, waiting for some more responses and reviews.

src/test/java/io/vertx/mqtt/test/server/MqttServerKeepAliveTest.java
vietj: it's really not useful here, Vert.x Unit will anyway consider the test is not finished until async.complete() is called even though the test method returns

vietj:
@ppatierno can you finish the review and merge for 3.8 / master ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -653,21 +653,46 @@ private void initChannel(ChannelPipeline pipeline) {
...
-            if (e.state() == IdleState.WRITER_IDLE) {
+            switch (e.state()) {
+            case READER_IDLE:
+              // verify that server is still connected (e.g. when using QoS-0)
               ping();
+              break;
+            case WRITER_IDLE:
+              // send ping or broker will close connection
+              ping();
+              break;
```
ppatierno: Can we do "fall through" here? Both cases have the same `ping()` instruction

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -653,21 +653,46 @@ private void initChannel(ChannelPipeline pipeline) {
...
             }
           }
         }
       });
+
+      if(this.options.getKeepAliveTimeout() > 0) {
```
ppatierno: Space after `if`

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -243,11 +243,10 @@ private void handleConnect(MqttConnectMessage msg) {
     if (msg.variableHeader().keepAliveTimeSeconds() != 0) {
 
       // the server waits for one and a half times the keep alive time period (MQTT spec)
-      int timeout = msg.variableHeader().keepAliveTimeSeconds() +
-        msg.variableHeader().keepAliveTimeSeconds() / 2;
+      int keepAliveTimeout =(int) (msg.variableHeader().keepAliveTimeSeconds() * 1.5);
```
ppatierno: space after `=`

src/test/java/io/vertx/mqtt/test/server/MqttServerKeepAliveTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * MQTT client keep alive tests using a Vert.x MQTT server to accomodate testing.
```
ppatierno: I guess this comment is wrong, the test is about MQTT server not client.

ppatierno:
@vietj  @konradmichael I have left a few comments but after them I think we are good to go.

vietj:
it seems the test did fail in CI

vietj:
tests still don't pass, we should have a CI passing to accept this PR - anybody wanting to contribute please step up and help

vietj:
I'm looking at finishing this PR and it seems that the client (which is visible in MqttClientKeepAliveTest) will send 2 pings simultaneously after 2 seconds because the IdleStateHandler will trigger for read and write. That seems incorrect to me.

vietj:
See https://github.com/vert-x3/vertx-mqtt/pull/173

Issue #128, opened 2019-03-28
Support binary last will message


paul-lysak:
At the moment it's only possible to use text Last Will messages because `MqttWill.getWillMessage` returns `String`.
MQTT brokers should also support binary messages, so it should be a byte array instead.
In fact, `MqttConnectPayload.willMessage` which is used to populate that field is deprecated in `netty-codec-mqtt`,
it's suggested to use `MqttConnectPayload.willMessageInBytes` instead.
Can we have an actual data stored as byte array in `MqttWill` and existing method `getWillMessage` turned into the wrapper
doing something like `new String(willMessage, CharsetUtil.UTF_8)`?

ppatierno:
@paul-lysak thanks for the PR, I left a comment there.

paul-lysak:
@ppatierno yes, I saw it and fixed.

PR #130, opened 2019-04-01
Binary Last Will message support

paul-lysak:
Fix for https://github.com/vert-x3/vertx-mqtt/issues/128 - adds support of binary body for the Last Will message, avoids using deprecated `MqttConnectPayload.willMessage` method.


src/main/java/io/vertx/mqtt/MqttWill.java
```diff
@@ -77,8 +79,16 @@ public String getWillTopic() {
 
   /**
    * @return the payload for the will as provided by the remote MQTT client
+   * @deprecated use {@link #getWillMessageBytes()} instead
    */
   public String getWillMessage() {
```
ppatierno: Add `@Deprecated` here, plese.

Issue #127, opened 2019-03-28
Ability to specify publish messageId manually


paul-lysak:
When using `vertx-mqtt` to implement a server with durable session storage (that survives server restarts) it's necessary to be able to manually
specify messageId when calling `MqttEndpoint.publish`. Currently the only option for assigning messageId is `MqttEndpointImpl.nextMessageId` which is purely memory-based
and if the server is restarted or migrated to another node it will start counting from the beginning, thus producing duplicate IDs.

PR #131, opened 2019-04-03
Explicit messageId for PUBLISH

paul-lysak:
Fix for https://github.com/vert-x3/vertx-mqtt/issues/127. Servers which retain session across the restarts can't rely on built-in `messageId` generation as it's purely in-memory and after the restart, it may generate the same `messageId`s as it generated before and which may not be fully processed yet. For such cases, we need to be able to provide `messageId` from some external source.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
@@ -145,7 +145,9 @@ public class MqttEndpointImpl implements MqttEndpoint {
     return this.protocolVersion;
   }
 
-  public String protocolName() { return this.protocolName; }
+  public String protocolName() {
+    return this.protocolName;
+  }
 
   public int keepAliveTimeSeconds() {
     return this.keepAliveTimeoutSeconds;
@@ -438,11 +440,19 @@ public class MqttEndpointImpl implements MqttEndpoint {
 
   @Override
   public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, Handler<AsyncResult<Integer>> publishSentHandler) {
+    return publish(topic, payload, qosLevel, isDup, isRetain, this.nextMessageId(), publishSentHandler);
+  }
+
+  @Override
+  public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, int messageId, Handler<AsyncResult<Integer>> publishSentHandler) {
+    if (messageId > MAX_MESSAGE_ID || messageId < 0) {
+      throw new IllegalArgumentException("messageId must be non-negative integer not larger than " + MAX_MESSAGE_ID);
+    }
 
     MqttFixedHeader fixedHeader =
       new MqttFixedHeader(MqttMessageType.PUBLISH, isDup, qosLevel, isRetain, 0);
     MqttPublishVariableHeader variableHeader =
-      new MqttPublishVariableHeader(topic, this.nextMessageId());
+      new MqttPublishVariableHeader(topic, messageId);
 
     ByteBuf buf = Unpooled.copiedBuffer(payload.getBytes());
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
```diff
@@ -438,11 +440,19 @@ public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel,
 
   @Override
   public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, Handler<AsyncResult<Integer>> publishSentHandler) {
+    return publish(topic, payload, qosLevel, isDup, isRetain, this.nextMessageId(), publishSentHandler);
+  }
+
+  @Override
+  public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, int messageId, Handler<AsyncResult<Integer>> publishSentHandler) {
+    if(messageId > this.messageIdCounter % MAX_MESSAGE_ID || messageId < 0) {
```
ppatierno: let's think about the case when the MQTT server crash but for handling sessions restore you have stored the latest messageId somewhere. On restarting the `this.messageIdCounte` will be zero so the first condition will not check that you are not setting a value not higher then MAX_MESSAGE_ID because the module will be always 0, right? The `messageIdCounter` is used only if you use the `nextMessageId` so we should not use it here or we should have a way for setting it and then after recovering you can continue to use the publish leveraging the `nextMessageId`. Wdyt? What's the case you are working on?

paul-lysak: Sorry, there was a dumb mistake in the range validation code. Please take a look at the updated version.

ppatierno: Yeah now it's better but I would like to know how you are going to use this feature ... I mean, you now have the possibility to set the `messageId` so are you going to use always this way or do you think that after recovering a session (from a storage) so sending all the pending messages ids, is useful to set the `messageIdCounter` and relying on the automatic `nextMessageId`? Just thinking aloud ... depending on the use case you have.

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
```diff
@@ -438,11 +440,19 @@ public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel,
 
   @Override
   public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, Handler<AsyncResult<Integer>> publishSentHandler) {
+    return publish(topic, payload, qosLevel, isDup, isRetain, this.nextMessageId(), publishSentHandler);
+  }
+
+  @Override
+  public MqttEndpointImpl publish(String topic, Buffer payload, MqttQoS qosLevel, boolean isDup, boolean isRetain, int messageId, Handler<AsyncResult<Integer>> publishSentHandler) {
+    if(messageId > MAX_MESSAGE_ID || messageId < 0) {
```
ppatierno: Just a couple of nits ...
Space after `if` and space before and after the `+` on string concatenation. Thanks!

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
paul-lysak: In my use case, messageId will always be specified manually, from a persistent external source. If you think that it might be useful - I can also change it to update `messageIdCounter` whenever `messageId` is provided manually.

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
paul-lysak: Sure, fixed that.

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
ppatierno: Well, I don't think that the counter should be update on publish because publishing messages could even happens not in the order of the message id. It could happen in some cases that messages are published with different QoS (1 or 2, where messageId makes sense) and someone want to just to procede to the next message if sending the current one fails somehow. It means that on retry, the message counter is set to an old previous message id in this case. Let's think more about that but for a different PR if needed.

PR #141, opened 2019-07-10
Server must send CONNACK 0x01 if CONNECT contains unsupported version…

### LINKED ISSUE #132: Server must send CONNACK 0x01 if CONNECT contains unsupported version of protocol
*Expected behavior:* in accordance with section 3.1.2.2 of MQTT Spec (http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html) "The Server MUST respond to the CONNECT Packet with a CONNACK return code 0x01 (unacceptable protocol level) and then disconnect the Client if the Protocol Level is not supported by the Server" 

*Actual behavior:* connection is closed by the server without sending CONNACK. Can be checked with `TC_MQTT_BROKER_CONNACK_005` test case from Eclipse IoT-Testware MQTT Test Suite (https://iottestware.readthedocs.io/en/master/mqtt_test_suite.html). 

*Investigation details:* actual check of protocol version is done by `io.netty.handler.codec.mqtt.MqttVersion.fromProtocolNameAndLevel` which throws `MqttUnacceptableProtocolVersionException`.
It's picked by `io.vertx.mqtt.impl.MqttServerConnection.handleMessage` as a failed `DecoderResult` and pushed further through `chctx.pipeline().fireExceptionCaught(result.cause());`.
Finally it gets to `io.vertx.core.net.impl.VertxHandler.exceptionCaught` which closes the channel. It appears to me that `MqttServerConnection.handleMessage` is the right place to give the 
`MqttUnacceptableProtocolVersionException` special treatment and send `CONNACK` prior to closing the connection.

vietj:
… of protocol - fixes #132

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -22,6 +22,7 @@ import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.codec.DecoderResult;
 import io.netty.handler.codec.mqtt.MqttConnectMessage;
 import io.netty.handler.codec.mqtt.MqttConnectReturnCode;
+import io.netty.handler.codec.mqtt.MqttUnacceptableProtocolVersionException;
 import io.netty.handler.timeout.IdleState;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.handler.timeout.IdleStateHandler;
@@ -87,7 +88,13 @@ public class MqttServerConnection {
 
       DecoderResult result = mqttMessage.decoderResult();
       if (result.isFailure()) {
-        chctx.pipeline().fireExceptionCaught(result.cause());
+        Throwable cause = result.cause();
+        if (cause instanceof MqttUnacceptableProtocolVersionException) {
+          endpoint = new MqttEndpointImpl(so, null, null, null, false, 0, null, 0);
+          endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_UNACCEPTABLE_PROTOCOL_VERSION);
+        } else {
+          chctx.pipeline().fireExceptionCaught(result.cause());
+        }
         return;
       }
       if (!result.isFinished()) {
--- a/src/test/java/io/vertx/mqtt/test/server/MqttServerSubscribeTest.java
+++ b/src/test/java/io/vertx/mqtt/test/server/MqttServerSubscribeTest.java
@@ -17,8 +17,10 @@
 package io.vertx.mqtt.test.server;
 
 import io.netty.handler.codec.mqtt.MqttQoS;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
+import io.vertx.core.net.NetClient;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.VertxUnitRunner;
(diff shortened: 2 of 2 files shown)


src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -87,7 +90,17 @@ void handleMessage(Object msg) {
...
+        if (cause instanceof MqttUnacceptableProtocolVersionException) {
+
+          MqttMessage connack = MqttEndpointImpl.createConnack(
+            MqttConnectReturnCode.CONNECTION_REFUSED_UNACCEPTABLE_PROTOCOL_VERSION,
+            false);
+          so.writeMessage(connack);
+          so.close();
+        } else {
+          chctx.pipeline().fireExceptionCaught(result.cause());
+        }
```
ppatierno: Thinking about this approach it's clear why you did that btw ...
The `connack` method in the `MqttEndpointImpl` already handles closing the connection if the returncode is not about connection accepted so we are duplicating the logic here.
It's also true that here we don't have an `MqttEndpoint` instance which is created later.
Anyway, in order to use the same logic, I would add a new ctor to the `MqttEndpointImpl` accepting just the `conn` parameter and then using the regular `connack` method on it to return the unacceptable protocol version error.
Wdyt?

vietj: I believe it cannot work because the message we receive is partially decoded as it is a `MqttMessage` instance that is not an  `MqttConnectMessage` instance (because it cannot be constructed with an invalid version protocol version) so we cannot really initialise the endpoint.

ppatierno: For this reason, I was talking about a new ctor where the `MqttEndpointImpl` is initialized just with the connection not with "will" information for example that we don't have of course.
Having an endpoint instance helps anyway to avoid code duplication.

vietj: to be honest this will be an artificial control flow to me, there is not much duplicated here

vietj: the code is pretty straightforward and is implemented according to the spec when the exception is actually available

vietj: I reworked the code to do like you said, my personal opinion is stil what I said above.

we need to get this merged soon for 3.8 release

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
```diff
@@ -122,6 +122,17 @@ public MqttEndpointImpl(NetSocketInternal conn, String clientIdentifier, MqttAut
...
+  public MqttEndpointImpl(NetSocketInternal conn) {
+    this.conn = conn;
+    this.clientIdentifier = null;
+    this.auth = null;
+    this.will = null;
+    this.isCleanSession = false;
+    this.protocolVersion = 0;
+    this.protocolName = null;
+    this.keepAliveTimeoutSeconds = 0;
+  }
```
ppatierno: why not calling the other ctor passing the above parameters?

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
```diff
@@ -295,14 +306,19 @@ public MqttEndpointImpl exceptionHandler(Handler<Throwable> handler) {
     }
   }
 
-  private MqttEndpointImpl connack(MqttConnectReturnCode returnCode, boolean sessionPresent) {
+  static io.netty.handler.codec.mqtt.MqttMessage createConnack(MqttConnectReturnCode returnCode, boolean sessionPresent) {
```
ppatierno: This method isn't needed anymore if we are going to create an `MqttEndpointImpl` instance so we can just call the `connack` method as before.

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -87,8 +90,13 @@ void handleMessage(Object msg) {
 
       DecoderResult result = mqttMessage.decoderResult();
       if (result.isFailure()) {
-        chctx.pipeline().fireExceptionCaught(result.cause());
-        return;
+        Throwable cause = result.cause();
+        if (cause instanceof MqttUnacceptableProtocolVersionException) {
+          //
```
ppatierno: my idea was about creating the endpoint here and sending the message not change the overall logic of handleConnect. Doesn't it work easily?

vietj:
@ppatierno updated according to review

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
ppatierno: I think we had a misunderstanding I meant just:

```java
Throwable cause = result.cause();
if (cause instanceof MqttUnacceptableProtocolVersionException) {
    endpoint = new MqttEndpointImpl(so);
    endpoint.reject(MqttConnectReturnCode.CONNECTION_REFUSED_UNACCEPTABLE_PROTOCOL_VERSION);
}
chctx.pipeline().fireExceptionCaught(result.cause());
return;
```
whitout changing anything else in the handleConnect but adding just a ctor for MqttEndpointImpl with so as parameter.

vietj: it's not clear what I should do perhaps you can modify yourself the PR and the right commit ?

PR #144, opened 2019-07-10
Client message QoS state improvements - fixes #143 - fixes #133

### LINKED ISSUE #133: mqttClient qos2inbound queue   keeps growing
When does the qos2inbound queue release when qos2?
image

I have an application that USES MQTT qos2 to send and receive messages and watch the memory usage grow.

The mat analysis is as follows:

image

### LINKED ISSUE #143: Client message QoS state improvements
Currently the client uses 3 `LinkedHashMap` to manage the message QoS.

- `qos1outbound` and `qos2outbound` can be replaced by `HashSet` as we don't need to retain the original message for later use
- `qos2inbound` can be replaced by `HashMap` as we don't need to retain any specific order

Both changes will improve GC and memory usage on the client.

In addition the `qos2inbound` as a leak (reported https://github.com/vert-x3/vertx-mqtt/issues/133) and it should remove from the map when the client receives the PUBREL message.

vietj:


vietj:
another review for you @ppatierno

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -110,14 +111,14 @@
   private Handler<Void> closeHandler;
 
   // storage of PUBLISH QoS=1 messages which was not responded with PUBACK
-  private LinkedHashMap<Integer, io.netty.handler.codec.mqtt.MqttMessage> qos1outbound = new LinkedHashMap<>();
+  private HashSet<Integer> qos1outbound = new HashSet<>();
 
   // storage of PUBLISH QoS=2 messages which was not responded with PUBREC
   // and PUBREL messages which was not responded with PUBCOMP
-  private LinkedHashMap<Integer, io.netty.handler.codec.mqtt.MqttMessage> qos2outbound = new LinkedHashMap<>();
+  private HashSet<Integer> qos2outbound = new HashSet<>();
```
ppatierno: the message was there because there is another PR related to restoring messages from storage when the client crashes and the inflight queue has to be rebuilt, so the client is able to resend messages to the server.

vietj: ok, we can keep the map then right ?

ppatierno: Yes. The PR should be just about avoiding the memory leak removing the message from the `qos2inbound`

vietj: ok, I'll keep though the usage of HashMap over LinkedHashMap that does not seem to be useful

vietj:
PR updated

ppatierno:
@vietj are you cherry-picking to master as well?

vietj:
yes I will soon

vietj:
done

PR #160, opened 2020-06-10
Handle missing connack message

### LINKED ISSUE #159: MQTT client dies if the broker does not send a CONNACK message
### Version

4.0.0-SNAPSHOT

### Context

I encountered an exception which looks suspicious while implementing the reconnect mechanism of our service. I'd like our verticle to retry the initial connection until the broker is up and running which works quiet well if the broker is fast enough. Otherwise, if the broker accepts the TCP connection but does not send a CONNACK message, the connectPromise() is not called and the verticle is stuck.

### Reproducer

I've made a simple repository to demonstrate the behaviour. 

* https://github.com/deen13/vertx-reconnect

### Steps to reproduce

1. Start the Verticle without having the broker up and running
2. Wait until the first connect attempt fails
3. Start the broker
4. Wait for the established connection 
5. Publish a message on the `test` topic

### Extra

The reconnect mechanism works if the retry interval is higher and the vernemq has some time to start. I wonder if this is a bug within this library or if the broker does not behave correctly.

deen13:
Fixes #159. I dont know if 20 seconds is "a reasonable" default amount of time because this depends on the environment.


vietj:
I think the option should de renamed because the proposal to use `connectionTimeout` is misleading with the existing TCP `connectTimeout`. As this means that the connection is established with an explicit server ACK, we should name it instead `connectionEstablishedTimeout`

PR #163, opened 2020-08-14
Fix execution of integration tests

sophokles73:
Running the integration tests via mvn verify failed because a standard
unit test has been executed as well which caused a bind address
conflict.

The test classes now use the standard naming pattern using *Test for
unit tests and *IT for integration tests. Accordingly, the surefire and
failsafe plugins do not need special include and exclude configuration
anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -118,9 +118,6 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
-          <excludes>
-            <exclude>**/MqttClient*.java</exclude>
-          </excludes>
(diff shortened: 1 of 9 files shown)


vietj:
could we rather move the it tests in an "it" package ?

sophokles73:
And keep the original naming? Then you will again need to configure includes and excludes for the surefire and failsafe plugins.
How about moving the integration tests into an IT package but still adopt the standard naming pattern?

sophokles73:
@vietj how about this?

sophokles73:
@vietj any additional thoughts?

vietj:
thanks @sophokles73 !

Issue #162, opened 2020-08-13
MQTT client never removes messages from outbound queue if not ACKed


sophokles73:
### Version

Which version(s) did you encounter this bug ?

3.9.2

### Context

When the MQTT client is used to publish a message using QoS 1 and the server does not send a PUBACK for the message, then the message will remain forever in the client's `qos1outbound` queue. If this happens repeatedly, the limit set by *maxInflightQueue* will eventually be exceeded and the client can no longer send any messages. 

The client itself (currently) does not keep track of pending PUBACKs by means of timers. Thus, it is the application level code's responsibility to do so. However, the client does not provide any means to remove a timed out PUBLISH attempt from the outbound queue once it determines that it won't wait any longer for a PUBACK. The client seems to have a similar problem with CONNECT as pointed out in #17 

I can see two options to address this issue:
1. Add support for timing out a PUBLISH request in the client itself or
2. Allow application code to declare a PUBLISH request as failed, thus removing it from the outbound queue.

I guess option 2 could be easily added without breaking backwards compatibility. The groundwork done as part of that could later be used to implement option 1 as well.

WDYT? I can create a PR for any or all of the options ...

vietj:
I think best option would be a timer that removes the pending message and fails the associated future is failed with a `TimeoutException`.

In addition we should provide two event handlers

- one for phantom acks (there is already a statement for this `log.warn("Received PUBACK packet without having related PUBLISH packet in storage")`)
- one for expired message

Both provide the opportunity to the application to save the state somewhere (like a persistent storage) and later check it with the other handler.

sophokles73:
> I think best option would be a timer that removes the pending message and fails the associated future is failed with a TimeoutException.

Sounds good.

> In addition we should provide two event handlers

We could do that by means of overloading the *publishCompletionHandler* method with new/different handlers, e.g.
```
MqttClient publishCompletionHandler(Handler<AsyncResult<Integer>> publishCompletionHandler, Handler<Integer> phantomAckHandler);
```
FMPOV we should then deprecate the original one. WDYT?

vietj:
I think it would be better another distinct handler

vietj:
that being said I think we cannot use AsyncResult<Integer> here instead we should have 3 handlers

- `publishCompletionHandler(Handler<Integer>)` (existing one)
- `publicCompletionExpirationHandler(Handler<Integer>)`
- `publishCompletionPhantomHandler(Handler<Integer>)`

[...]

PR #165, opened 2020-08-25
[#162] Add support for handling timeouts of expected packets

sophokles73:
The MQTT client now supports setting a timeout value and handlers that
will be invoked when an expected packet from the server in response to a
QoS 1 or 2 PUBLISH packet does not arrive in time.

If a response times out, the corresponding PUBLISH packet is removed
from the outbound queue in order to prevent blocking subsequent PUBLISH
messages being via the client.


src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -294,6 +297,36 @@ public MqttClientOptions setKeepAliveTimeSeconds(int keepAliveTimeSeconds) {
     return this;
   }
 
+  /**
+   * Sets the time after which the client stops waiting for a PUBACK or PUBREC
```
vietj, line 301: _Sets the time in seconds_

vietj:
@ppatierno can you review this contribution of a new feature and let us know if that's fine ?

src/main/java/io/vertx/mqtt/MqttClientOptions.java
sophokles73: The unit is already defined in the parameter description by I can, of course, put it here as well ...

ppatierno:
@sophokles73 out of curiosity I have a couple of questions ...

1. I haven't found any reference in the documentation about not waiting for an ack anymore after a timeout so that the right behavior (even if reasonable) is to remove the PUBLISH from the outbound queue and not waiting for the ack anymore. Did I miss something?
2. You mentioned, "to prevent blocking subsequent PUBLISH messages being via the client.". It should happen only if the inflight queue is full but not in normal conditions. Is this the case you are referring?

ppatierno:
I found the answer to 2) reading #162 (sorry just catching up with notifications after 2 weeks of vacation :-))
Still curious about question 1).

sophokles73:
@ppatierno 

> I haven't found any reference in the documentation about not waiting for an ack anymore after a timeout so that the right behavior (even if reasonable) is to remove the PUBLISH from the outbound queue and not waiting for the ack anymore. Did I miss something?

Hi Paolo, I think I do not get the question here. Can you elaborate a little more?

ppatierno:
@sophokles73 I meant we are adding behavior that is not described in the spec. I was wondering if it's the right thing that people using the component would expect.

sophokles73:
@ppatierno My understanding is that for a client it doesn't make much sense to wait forever for a PUBACK. Instead, after a reasonable amount of time, it basically has two options:
1. consider the PUBLISH failed
2. re-send the packet using the same packet ID and wait again for the PUBACK

The newly added handlers and the timeout should provide means for client code to implement any of the two options. Without these handlers, the client becomes stale/useless if the receiver doesn't send a PUBACK or PUBREC at all.

The default behavior could be for the client to wait forever (timeout = -1). But if a timeout and a handler is set, then FMPOV it is reasonable to remove the failed/timed-out message from the outbound queue after we have notified client code about the timeout via the handlers. WDYT?

src/main/java/io/vertx/mqtt/MqttClient.java
```diff
@@ -142,6 +142,32 @@ static MqttClient create(Vertx vertx) {
...
+
+  /**
+   * Sets a handler which will be called when the client receives a PUBACK/PUBREC/PUBCOMP with an unknown
+   * packet ID.
+   *
+   * @param publishCompletionPhantomHandler the handler to call with the unknown packet ID
+   * @return current MQTT client instance
+   */
+  @Fluent
+  MqttClient publishCompletionPhantomHandler(Handler<Integer> publishCompletionPhantomHandler);
```
ppatierno, line 169: While not using a name with "unknown" instead of this "phantom" thing?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -1061,4 +1180,18 @@ private boolean isValidStringSizeInUTF8(String string) {
 
     return false;
   }
+
+  private class ExpiringPacket {
```
ppatierno, line 1184: Can you add a little bit of Javadoc for the class, please?

ppatierno (approved):
LGTM, just a couple of nits.

vietj:
I think it is fine as long as it is an optional and documented feature (i.e
expliciting in the doc that doing that is a commodity that might not follow
the spec)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
vietj: @sophokles73 ping

src/main/java/io/vertx/mqtt/MqttClient.java
vietj: agreeing with you @ppatierno - @sophokles73 ping

vietj:
@sophokles73 a couple of changes and we can merge this PR

sophokles73:
I will make the changes this afternoon. Would that be still in time for 3.9.3?

src/main/java/io/vertx/mqtt/MqttClient.java
sophokles73: @vietj  how about *publishCompletionUnknownPacketIdHandler*?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
sophokles73: will do

vietj:
yes we are releasing 3.9.3 tomorrow.

sophokles73:
@vietj  I have pushed changes ...

sophokles73:
@vietj I think that I have missed something: when a PUBLISH times out and the corresponding handler is invoked with the packet ID, then one option for the client code should be to re-try the sending of the PACKET with the same packet ID. However, there currently is no way for the client code to do so. Invoking the publish method always assigns a new packet ID but the MQTT spec mandates to use the same packet ID in this case if I am not mistaken. WDYT @ppatierno ?

sophokles73:
I can imagine doing the following:
If *ackTimeout* is set to a value > 0, then the client will re-send a PUBLISH packet up to a configurable number of times, if no publishCompletionTimeoutHandler has been set. Otherwise, the handler is invoked if the PUBLISH times out. Only in the latter case the published packet will be removed from the outbound queue.

Another option would be to add methods *cancelPublish(int)* and something like *republish(int)* to allow client code to decide whether to consider the publishing failed or to re-try when the publishCompletionTimeoutHandler has been invoked.

FMPOV the first option seems more appropriate. WDYT @vietj @ppatierno ?

vietj:
@sophokles73 does the MQTT spec mentions retries ? I believe that MQTT provides reliability settings and uses TCP as transport which is reliable, so the server should received the packet and I don't think there is a need to retry. If you don't get an answer then it's a server failure and the client should not retry. WDYT @ppatierno ?

vietj:
can we merge this branch in the meantime ?

sophokles73:
@vietj FMPOV the MQTT 3.1.1 spec is not really clear about whether a client is expected/allowed to re-send a packet while still being connected to the server. [Section 2.3.1](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718025) explicitly mentions that
> If a Client re-sends a particular Control Packet, then it MUST use the same Packet Identifier in subsequent re-sends of that packet.

The spec also states in its abstract
> The protocol runs over TCP/IP, or over other network protocols that provide ordered, lossless, bi-directional connections.

So I guess you are right in assuming that the server's acknowledgement should not get lost if implemented correctly and the connection is still established.

I guess we might still add re-send behaviour in another step (PR) later. So I think we can merge this PR for the time being ...

vietj:
yes a separate PR would be better as we can ship this PR in 3.9

vietj:
@sophokles73 can you back port this to 3.9 branch with a new PR ? the test you added uses Mockito and vertx internals exclusive to Vert.x 4.0 and thus make the back port non trivial

sophokles73:
@vietj ok, I'll give it a shot ...

sophokles73:
@vietj I have created #171

sophokles73:
@vietj can you also please include #172 in 3.9.3?

vietj:
it seems fine @sophokles73 https://github.com/vert-x3/vertx-mqtt/blob/3.9/src/test/java/io/vertx/mqtt/impl/MqttClientImplTest.java#L2

Issue #164, opened 2020-08-20
Possible memory leak when using many MqttClient instances one after another


mbaeuerle:
### Questions

-

### Version

Which version(s) did you encounter this bug ?
3.9.1

### Context

The setup is as follows:
A Vert.X MqttClient is created, connecting, sending/receiving some data and then closing the connection again.
This is done multiple times per minute each time a new MqttClient is created, overwriting the last one.
It seems after ca 2h the JVM uses 200-300MB more memory than before and increases even more.

### Do you have a reproducer?

Unfortunately not.

### Steps to reproduce

1. Create MqttClient
2. Connect to Broker
3. Receive some message
4. Disconnect
5. Start again at 1

### Extra

* Using VisualVM I looked at the diff between two snapshots some hours apart. There it looked like during this time many additional (in this case over 700) `NetClientImpl` objects were allocated but no new `MqttClientImpl`.
* When checking the code I noticed that the `MqttClientImpl` is connecting via the `NetClient`:
  https://github.com/vert-x3/vertx-mqtt/blob/d51c272d4de5dd148843f5a6b30bfadfaee26185/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java#L195
  But disconnection is not done by calling `client.close()` but by closing the connection directly:
  https://github.com/vert-x3/vertx-mqtt/blob/d51c272d4de5dd148843f5a6b30bfadfaee26185/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java#L275
  Maybe this misses some of the clean up steps which are done when calling `NetClientImpl.close()`:
  https://github.com/eclipse-vertx/vert.x/blob/581ce8d8ff55652269030d20d9fc95b0ac1b1808/src/main/java/io/vertx/core/net/impl/NetClientImpl.java#L120-L133

vietj:
can you perform a heap dump when this happens and post it somewhere ?

sophokles73:
Here's a Gist with a reproducer: https://gist.github.com/sophokles73/84965340c3098478a6a2a1164cecbb67
And a profiler snapshot that can be opened with VisualVM. The snapshot has been taken after about 2500 connect/dsiconnect cycles using the gist above.
[vertx-mqtt-client-snapshot-1598274789602.zip](https://github.com/vert-x3/vertx-mqtt/files/5118026/vertx-mqtt-client-snapshot-1598274789602.zip)

vietj:
thanks

[...]

PR #166, opened 2020-08-25
[#164] Prevent leaking of NetClient instances

sophokles73:
The NetClientImpl used by MqttClientImpl to connect to a server
requires its close() method to be invoked to release some resources
which would otherwise prevent it from being GCed properly.

The MqttClientImpl's disconnect() method has been changed to invoke the
NetClientImpl.close() method instead of just the close() method of the
NetSocketImpl instance that represents the connection to the MQTT
server.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -270,11 +270,13 @@ public class MqttClientImpl implements MqttClient {
 
     io.netty.handler.codec.mqtt.MqttMessage disconnect = MqttMessageFactory.newMessage(fixedHeader, null, null);
 
-    this.write(disconnect);
-
-    connection().close();
-
-    return ctx.succeededFuture();
+    Promise<Void> result = Promise.promise();
+    this.write(disconnect)
+      .onComplete(r -> {
+        this.connection = null;
+        this.client.close(result);
+      });
+    return result.future();
   }
 
   /**


vietj:
thanks, in master we might do differently as we changed the lifecycle for clients

vietj:
you can have a look at latest SQL client that creates an un-managed `NetClient` (using `VertxInternal)` and registers a close hook for itself. We should do the same for MQTT client in master.

vietj:
This shows how the a client can register a close hook https://github.com/eclipse-vertx/vertx-sql-client/blob/master/vertx-pg-client/src/main/java/io/vertx/pgclient/impl/PgPoolImpl.java#L52

vietj:
I think however for 3.9 we should simply do as you proceeded

sophokles73:
@vietj are you proposing to create two separate PRs, one for master and one for 3.9?

vietj:
indeed, both versions won't address client closing the same way

sophokles73:
@vietj The MqttClientImpl allows the same client instance to be used to re-connect to a server after its disconnect() method has been invoked. This behavior is not explicitly specified by the MqttClient JavaDocs but closing the underlying NetClientImpl as part of the disconnect() method would prevent client code from taking advantage of this behavior. I therefore suggest to add a *close()* method to the *MqttClient* interface instead, which closes the connection *and* the *NetClientImpl*. Client code can then decide if it wants to re-use an MqttClient instance or not. Using the *close()* method would then make sure that the discarded MqttClient can be cleany GCed. WDYT?

vietj:
I think it's good

ppatierno:
@sophokles73 are you going to do the latest changes you mentioned on this PR or on a new one?

sophokles73:
@ppatierno yes, that is my intention. I also think that we can use the same approach for both master and 3.9. WDYT?

vietj:
3.9 does not have a CloseFuture and it will not work the same way. However in 3.9 you can also implement the automatic close if that's used within a Verticle, the code will be slighly different and you cannot really backport the commit.

vietj:
I'll handle this myself in master and back port this change as is in 3.9 for 3.9.3

vietj:
So actually I think what is missing is a close() operation on the MQTT client that would close the underlying NetClient.

Closing the NetClient on the MQTT client will prevent reconnecting unless we make the client not final anymore and create an instance when it's needed.

vietj:
I went in a slight different way, since we are using a single connection per client, we instantiate the client when connect() is called. When the connection is closed, the net client is closed too. I've changed to set the ctx and connection field to null when the close handler is called. See https://github.com/vert-x3/vertx-mqtt/pull/169

vietj:
See https://github.com/vert-x3/vertx-mqtt/pull/170

PR #167, opened 2020-08-30
Use ephemeral ports for MQTT server tests

sophokles73:
This allows for running multiple test classes in parallel.


vietj:
@sophokles73 I would like to avoid that because it makes debugging with wireshark harder.

sophokles73:
I see

PR #169, opened 2020-09-14
Instantiate the NetClient when client connection happens, when the Ne…

vietj:
…tSocket close handler is called then close the client and the connection + context to null


vietj:
Can you have a quick look @ppatierno ?

PR #225, opened 2022-08-01
Add websocket compression with "sec-websocket-Extensions: permessage-deflate"

henm:
This pull request adds support for `permessage-deflate` compression to the mqtt-server if connected via a websocket.

The implementation and configuration is an adjusted version of the implementation in vert.x-core, see [here](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/http/impl/HttpServerConnectionHandler.java#L87) and [here](https://github.com/eclipse-vertx/vert.x/blob/master/src/main/java/io/vertx/core/http/HttpServerOptions.java#L120).


vietj:
thanks for the contribution

PR #231, opened 2022-11-18
Make default max message size the same as Netty.

mattisonchao:
## Motivation:

Make the default max message size the same as Netty `8092`, it can also help to support `maxClientIdLength` easily.

Snapshot:

image


mattisonchao:
@vietj  PTAL

mattisonchao:
I signed the ECA, the CI still fail though... 

image

PR #236, opened 2023-02-14
The default client id length is 23 it should be configurable instead

### LINKED ISSUE #230: add MAX_CLIENT_ID_LENGTH in MqttServerOptions
### Questions
 the  length of clientId can not bigger than 23 in the default MqttServer
### Version

4.3.5

### Context

when I use MqttServer, the  length of clientId can not bigger than 23

### Do you have a reproducer?

 I hope you can add the config of  MAX_CLIENT_ID_LENGTH  or expose the getMethod of  NetServer instance

vietj:
fixes #230


ppatierno:
@vietj I was going to review ... but it seems too late :-)

PR #244, opened 2023-10-20
convert will message in mqtt options from string to buffer

JeffreyThijs:
Motivation:

The payload of the will message now requires to be stored in the `MqttClientOptions` as a String and is afterwards converted to byte[] assuming UTF-8 in the `MqttClientImpl` when connecting. This is different from the publish methods where a `Buffer` is required. Additionally, when the will message payload is binary (e.g. in SparkplugB) converting this to a string feels wrong and might lead to unnecessary conversion issues.

### DIFF
--- a/src/main/java/io/vertx/mqtt/MqttClientOptions.java
+++ b/src/main/java/io/vertx/mqtt/MqttClientOptions.java
@@ -214,6 +214,7 @@ public class MqttClientOptions extends NetClientOptions {
    * @return will message content
    */
   @Deprecated
+  @GenIgnore
   public String getWillMessage() {
     return willMessage.toString(StandardCharsets.UTF_8);
   }


tsegismont:
Thanks for sharing this. Can you post some more details (e.g. snippet) about the problem?

src/main/generated/io/vertx/mqtt/MqttClientOptionsConverter.java
```diff
@@ -65,6 +65,8 @@ static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, MqttCli
             obj.setWillTopic((String)member.getValue());
           }
           break;
+        case "willMessageAsString":
+          break;
```
ppatierno, line 69: why is this generated and empty here? @vietj ?

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -202,10 +206,25 @@ public String getWillTopic() {
...
+  /**
+   * @return will message content as string
+   */
+  public String getWillMessageAsString() {
+    return getWillMessageAsString(StandardCharsets.UTF_8);
+  }
+
+  /**
+   * @return will message content as string
+   */
```
ppatierno, line 223: I would add the `@param` in the javadoc to make the difference from the other overload.

JeffreyThijs:
> Thanks for sharing this. Can you post some more details (e.g. snippet) about the problem?

Here is a reproducer: 

https://github.com/JeffreyThijs/vertx-mqtt-reproducer/blob/master/src/test/kotlin/jthijs/TestMqttWillMessage.kt

src/main/java/io/vertx/mqtt/MqttClientOptions.java
JeffreyThijs: I have added this

tsegismont (commented):
Thank you @JeffreyThijs for sharing this.

Can you please make the following changes in the client options class:

* deprecate `willMessage` string property
* add `willMessageBytes` buffer property
* use a buffer as the backing field for both properties

When the PR is ready, I will backport the changes to the `4.x` branch.

And eventually, remove the deprecated methods in the main branch.

src/main/generated/io/vertx/mqtt/MqttClientOptionsConverter.java
JeffreyThijs: should be fixed

JeffreyThijs:
@tsegismont, i think i made all changes you suggested.

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -202,7 +204,16 @@ public String getWillTopic() {
...
    */
+  @Deprecated
   public String getWillMessage() {
+    return willMessage.toString(StandardCharsets.UTF_8);
+  }
+
+  /**
+   * @return will message bytes content
+   */
+  @GenIgnore
```
tsegismont, line 215: Please mark the `willMessage` property with `@GenIgnore` instead of this one.

Then in the constructor which takes a `JsonObject`, add some code to take into account `willMessage` field, if present (transform to Buffer and log warning)

tsegismont (commented):
Thank you for the quick turnaround @JeffreyThijs

src/main/java/io/vertx/mqtt/MqttClientOptions.java
JeffreyThijs: Ok, i made these changed. However, https://github.com/vert-x3/vertx-mqtt/pull/244#discussion_r1372743872 is now reintroduced.

src/main/generated/io/vertx/mqtt/MqttClientOptionsConverter.java
ppatierno: @tsegismont is it normal having that empty one?

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -95,6 +102,11 @@ public MqttClientOptions(JsonObject json) {
     super(json);
     init();
     MqttClientOptionsConverter.fromJson(json, this);
+    if (json.containsKey("willMessage")) {
+      String willMessageString = json.getString("willMessage");
+      willMessage = Buffer.buffer(willMessageString);
+      log.warn(String.format("converting json value willMessage: %s to buffer", willMessageString));
```
tsegismont, line 108: Better safe than sorry, let's not log the value itself

tsegismont (commented):
@JeffreyThijs can you accept my suggestions and rebuild the project to see if the converter problem goes away?

src/main/generated/io/vertx/mqtt/MqttClientOptionsConverter.java
tsegismont: I think this can be due to the usage of `@GenIgnore`, I made some suggestions to @JeffreyThijs , let's see the result

tsegismont:
@JeffreyThijs @ppatierno I've made some updates and merged the PR

Thank you both!

tsegismont:
I will take care of backporting to `4.x`

PR #261, opened 2026-01-25
Pause/Resume read from Socket - branch 4.x

dometec:
PR for the old #6 :)

I want to pause reading from the socket to avoid being overwhelmed by QoS 0 messages.
When reading is paused, incoming data accumulates in the TCP receive buffer. As the buffer fills up, the TCP stack advertises a reduced window size (eventually a zero window) to the broker, applying backpressure at the TCP level.

At that point, outgoing data may accumulate in the broker’s socket send buffer until its limits are reached. Depending on the broker implementation and configuration, QoS 0 messages that cannot be written may be buffered or dropped.

### DIFF
--- a/src/main/java/io/vertx/mqtt/MqttClient.java
+++ b/src/main/java/io/vertx/mqtt/MqttClient.java
@@ -322,6 +322,29 @@ public interface MqttClient {
   MqttClient ping();
 
   /**
+   * Pause the reading channel, so no new byte are read from the server.
+   * Available onliy after connection is established.
+   * <p>
+   * This simply delegates to {@link io.vertx.core.net.NetSocket#pause()}.
+   */
+  void pause();
+
+  /**
+   * Returns true if the reading channel is paused.
+   * Available onliy after connection is established.
+   * @return true if the client publishing is paused.
+   */
+  boolean isPaused();
+
+  /**
+   * Resume the reading channel. see {@link #pause()}
+   * Available onliy after connection is established.
+   * <p>
+   * This simply delegates to {@link io.vertx.core.net.NetSocket#resume()}.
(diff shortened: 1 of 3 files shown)


vietj:
I think this should use instead NetSocket pause / resume instead

dometec:
It doesn’t have the same effect, in fact, it has no effect at all.
NetSocket.pause() switches the internal InboundBuffer into fetch mode, but this buffer is never actually used because MqttClientImpl sets a messageHandler instead of a handler, effectively bypassing the InboundBuffer backpressure mechanism.

I tried to simply replace messageHandler with handler here:
https://github.com/vert-x3/vertx-mqtt/blob/cc90db35b19a9b378100dcaf09acb4afd5001426/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java#L252

but doing so resulted in no messages being delivered at all.

If you think this is the correct direction, let me know and I’ll spend some more time understanding where the missing piece is.

vietj:
right, this is because messages are not accounted by back pressure and I think only works in master.

can you make a PR for master and use pause/resume so we check that ?

I would like to determine what is the long term viable option for this.

dometec:
on master it works with NetSocket, see PR https://github.com/vert-x3/vertx-mqtt/pull/262

Let me know how to proceed for this PR on 4.x branch

vietj:
the tests seem to do very large pause with the timers, is that possible to have a test that takes less time ?

dometec:
pause reduced, like for 5.0 branch, I had to update testcontainer to pass the CI

vietj:
thanks @dometec can you provide a PR for master that also upgrades the test ?

PR #265, opened 2026-03-13
MQTT Client V5 Implementation Vertx5

dometec:
This PR introduces initial MQTT v5 client support. The implementation is still in progress and will be updated incrementally to allow early feedback.

Since MQTT v5 extends LWT with additional configuration options, a dedicated class was introduced to keep the API clean and maintainable.

Some new methods in MqttClient are annotated with @GenIgnore because they expose Netty MqttProperties. As polyglot support appears to be limited to JVM languages now, this should be acceptable. Please let me know if a dedicated Vert.x data object is preferred instead.

Current progress

Done:
- CONNECT v5 with properties and LWT options
- CONNACK
- SUBSCRIBE
- SUBACK
- UNSUBSCRIBE
- UNSUBACK
- DISCONNECT

Todo (review, test, ecc.):
- PUBLISH
- PUBACK/PUBREC/PUBREL/PUBCOMP
- AUTH

Basic tests are included.


dometec:
Done, I want to do some tests with other brokers, but the implementation is complete.

PR #267, opened 2026-04-18
Fix Inverted condition on isPaused method

dometec:
#261 (Pause/Resume read from Socket - branch 4.x) introduce a bug in https://github.com/vert-x3/vertx-mqtt/blob/acb3c437b790bf2e2cddf57c95c62c2ff3a02f0c/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java#L733-L734 
The condition must be negated.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
@@ -731,7 +731,7 @@ public class MqttClientImpl implements MqttClient {
   }
 
   public synchronized boolean isPaused() {
-    return connOption.isAutoRead();
+    return !connOption.isAutoRead();
   }
 
   @Override


dometec:
Add recvByteBufAllocatorSize to MqttClientOptions and update converter.

ppatierno:
> Add recvByteBufAllocatorSize to MqttClientOptions and update converter.

I think PRs should be focused on just one thing.

dometec:
Sure, but this is related to always to pause/resume features as described in https://github.com/smallrye/smallrye-reactive-messaging/pull/3378

ppatierno:
How the code you changed around `!connOption.isAutoRead();` is impacting or is impacted by the receiver buffer size. It's not clear from the PR. Also the link you are referring to in smallrye seems to be more about the receiver buffer size than what this PR seems to aim to fix from the title "FIx Inverted condition on isPaused method".

dometec:
clear, let me move the option in a new PR.

ppatierno:
> Related to feature added in https://github.com/vert-x3/vertx-mqtt/pull/261

Sorry to be pedantic but I guess the description doesn't explain the changes. I think you should mention that the #261 introduced a bug whre the `isPaused` is wrongly defined and this PR is addressing such a bug.

dometec:
No Paolo, things should be done properly because many people will read this in the future. Thanks for your patience. If you agree, we can delete all comments not related to the PR just to reduce the noise.

ppatierno (approved):

PR #268, opened 2026-04-20
Add recvByteBufAllocatorSize to MqttClientOptions and update converter

dometec:
I want the ability to set RecvByteBufAllocatorSize to control how many bytes the  MQTT Decoder can process (to create MQTT Messages) after pausing the reads (autoread #261).

See also https://github.com/smallrye/smallrye-reactive-messaging/pull/3378


ppatierno (approved):
LGTM. Thanks!

PR #269, opened 2026-04-21
MQTT Client V5 Implementation (4.x branch)

dometec:
This is a backport of #265 to the 4.x branch.
The tests I’ve run so far are OK, but I’m leaving this as a draft (like #265) because I want to run additional.


.github/workflows/ci-client-mqtt5.yml
```diff
@@ -0,0 +1,26 @@
+name: CI client_mqtt5_vertx4
```
ozangunalp: Let's remove this workflow

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -521,6 +581,118 @@ public int getIdleTimeout() {
...
+    }
+    this.maximumPacketSize = maximumPacketSize;
+  }
+
+  public Integer getTopicAliasMaximum() {
+    return topicAliasMaximum;
+  }
+
+  public void setTopicAliasMaximum(Integer topicAliasMaximum) {
+    if (receiveMaximum != null &&
```
ozangunalp: This needs to check `topicAliasMaximum`

src/main/java/io/vertx/mqtt/MqttClientOptions.java
```diff
@@ -521,6 +581,118 @@ public int getIdleTimeout() {
...
+    }
+    this.receiveMaximum = receiveMaximum;
+  }
+
+  public Long getMaximumPacketSize() {
+    return maximumPacketSize;
+  }
+
+  public void setMaximumPacketSize(Long maximumPacketSize) {
+    if (sessionExpireInterval != null &&
```
ozangunalp: This needs to check `maximumPacketSize`

src/main/java/io/vertx/mqtt/MqttClientOptions.java
ozangunalp, line 497: We could make this return `MqttClientOptions` for builder chaining.

src/main/java/io/vertx/mqtt/MqttClient.java
```diff
@@ -17,17 +17,31 @@
...
 import io.vertx.mqtt.impl.MqttClientImpl;
 import io.vertx.mqtt.messages.MqttConnAckMessage;
+import io.vertx.mqtt.messages.MqttDisconnectMessage;
+import io.vertx.mqtt.messages.MqttPubAckMessage;
+import io.vertx.mqtt.messages.MqttPubCompMessage;
+import io.vertx.mqtt.messages.MqttPubRecMessage;
 import io.vertx.mqtt.messages.MqttPublishMessage;
 import io.vertx.mqtt.messages.MqttSubAckMessage;
+import io.vertx.mqtt.messages.MqttUnsubAckMessage;
+import io.netty.handler.codec.mqtt.MqttProperties;
```
ozangunalp: We need to move this import next to other netty imports.

src/main/java/io/vertx/mqtt/MqttClient.java
```diff
@@ -17,17 +17,31 @@
 package io.vertx.mqtt;
 
 import io.netty.handler.codec.mqtt.MqttQoS;
+import io.netty.handler.codec.mqtt.MqttSubscriptionOption;
```
ozangunalp, line 21: I was thinking that we should not expose Netty types from the Client API. 
But than `MqttQoS` is already there and`MqttProperties` is quite type-checked. 
It would not be very practical for the user to pass through custom types, or `JsonObject` as properties.

@vietj wdyt ?

src/main/java/io/vertx/mqtt/MqttClientWillOptions.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Set the QoS level for the will message.
+   *
+   * @param qos QoS level (0, 1 or 2)
+   * @return this options instance
+   */
+  public MqttClientWillOptions setQos(int qos) {
+    this.qos = qos;
```
ozangunalp, line 153: We can validate the qos input here.

src/main/java/io/vertx/mqtt/messages/impl/MqttAuthenticationExchangeMessageImpl.java
```diff
@@ -0,0 +1,47 @@
...
+  public MqttAuthenticateReasonCode reasonCode() {
+    return reasonCode;
+  }
+
+  @Override
+  @SuppressWarnings("rawtypes")
+  public String authenticationMethod() {
+    MqttProperties.MqttProperty prop =
+      properties.getProperty(MqttProperties.MqttPropertyType.AUTHENTICATION_METHOD.value());
+    return prop == null ? null : ((MqttProperties.StringProperty) prop).value();
```
ozangunalp, line 32: Let's check the type before casting to `StringProperty`.

src/main/java/io/vertx/mqtt/messages/impl/MqttAuthenticationExchangeMessageImpl.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  @SuppressWarnings("rawtypes")
+  public Buffer authenticationData() {
+    MqttProperties.MqttProperty prop =
+      properties.getProperty(MqttProperties.MqttPropertyType.AUTHENTICATION_DATA.value());
+    return prop == null ? null : Buffer.buffer(((MqttProperties.BinaryProperty) prop).value());
```
ozangunalp, line 40: Same here, let's check before casting to `BinaryProperty`

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
```diff
@@ -548,8 +549,8 @@ public MqttEndpointImpl publishComplete(int publishMessageId, MqttPubCompReasonC
 
     MqttFixedHeader fixedHeader =
       new MqttFixedHeader(MqttMessageType.PUBCOMP, false, MqttQoS.AT_MOST_ONCE, false, 0);
-    MqttMessageIdAndPropertiesVariableHeader variableHeader =
-      new MqttMessageIdAndPropertiesVariableHeader(publishMessageId, properties);
+    MqttPubReplyMessageVariableHeader variableHeader =
```
ozangunalp, line 552: I guess this was wrong before?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -967,60 +1461,137 @@ private void handleMessage(ChannelHandlerContext chctx, Object msg) {
...
           break;
 
         case UNSUBACK:
-          handleUnsuback(((MqttMessageIdVariableHeader) mqttMessage.variableHeader()).messageId());
+          handleUnsuback(mqttMessage);
           break;
 
         case PINGRESP:
           handlePingresp();
           break;
```
ozangunalp, line 1598: Shall we also have a handling of `AUTH` packet ? Like calling a handler `Handler<MqttAuthenticationExchangeMessage> authHandler` ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
ozangunalp, line 1129: what if the ChannelConfig `connOption` is not yet set ? Shall we accept pause/resume after we receive ConnAck ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -1329,6 +2073,67 @@ private String generateRandomClientId() {
...
+   * Parses a SERVER_REFERENCE string (comma-separated "host:port" or "host" entries)
+   * and returns {host, port} for a randomly chosen entry.
+   * IPv6 addresses enclosed in brackets are supported (e.g. "[::1]:1883").
+   *
+   * @return String array {host, portString}, or {@code null} if the string is blank/unparseable
+   */
+  private String[] pickServer(String serverReference) {
+    if (serverReference == null || serverReference.trim().isEmpty()) return null;
+    String[] entries = serverReference.split(",");
+    String entry = entries[ThreadLocalRandom.current().nextInt(entries.length)].trim();
```
ozangunalp, line 2140: @vietj would you have any guidance to avoid the usage of `ThreadLocalRandom` ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
```diff
@@ -1090,7 +1691,7 @@ private void handlePubackTimeout(int packetId) {
         // the message has already been ACKed
         log.debug("PUBLISH expiration timer fired but QoS 1 message has already been PUBACKed by server");
         return;
-     }
+      }
     }
     countInflightQueue--;
```
ozangunalp: I am not sure if calls to ++ or -- `countInflightQueue` are serialized. So should this be inside the synchronized block ?

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
ozangunalp: Same here, should this be inside the synch block ?

ozangunalp: Same here, should this be in synch block ?

ozangunalp (changes requested):
@dometec I've pointed out some flagrant things I saw about correctness.

ozangunalp:
@dometec would you have time to also open the PR targeting Vert.x 5 (master) ?

src/main/java/io/vertx/mqtt/MqttClientOptions.java
dometec: sure!

src/main/java/io/vertx/mqtt/messages/impl/MqttAuthenticationExchangeMessageImpl.java
dometec: I don't believe it's necessary, since the AUTHENTICATION_METHOD si a UTF8 String as spec and netty codec (codec-mqtt/src/main/java/io/netty/handler/codec/mqtt/MqttDecoder.java) create a MqttProperties.StringProperty for it. But let me known

src/main/java/io/vertx/mqtt/messages/impl/MqttAuthenticationExchangeMessageImpl.java
dometec: As before, let me known

src/main/java/io/vertx/mqtt/messages/impl/MqttAuthenticationExchangeMessageImpl.java
ozangunalp: OK for me

src/main/java/io/vertx/mqtt/impl/MqttEndpointImpl.java
dometec: yes, the reasonCode was lost

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
dometec: connOption is set at connection time (as connection), moreover that pause is called for sure after the client get some messages

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
dometec: good point, I've moved the two countInflightQueue access into synchronized block (the others nine access are already inner a synchronized block)

src/main/java/io/vertx/mqtt/impl/MqttClientImpl.java
dometec: I implemented the handler, let me try with EMQX (I use VerneMQ but it doesn't support AUTH packet) and I'll commit the changes

dometec: done!

PR #273, opened 2026-05-05
Add recvByteBufAllocatorSize to MqttClientOptions and update converter (master branch)

dometec:
This option is needed to control how many byte are going to be transformed in MQTT messages when the stream read is paused (see also https://github.com/smallrye/smallrye-reactive-messaging/pull/3378)


ppatierno (approved):
LGTM.

PR #276, opened 2026-05-05
WillTopic is required if WillFlag is true (master branch)

dometec:
I'm going to implement a check implementing the client MQTTv5 that verify the presente of WillTopic is WillFlag is true

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -229,15 +229,15 @@ public class MqttServerConnection {
     if (msg.variableHeader().isWillFlag()) {
       // [MQTT-3.1.2-9] if Will Flag is 1, Will Topic MUST be present in the Payload
       String willTopic = msg.payload().willTopic();
-      willInvariantViolated = willTopic == null || willTopic.isEmpty();
+      willInvariantViolated = willTopic == null
+        || willTopic.isEmpty()
+        || msg.payload().willMessageInBytes() == null;
     } else {
-      // [MQTT-3.1.2-11] if Will Flag is 0, Will QoS and Will Retain MUST be 0,
-      // and Will Topic / Will Message MUST NOT be present in the Payload
-      String willTopic = msg.payload().willTopic();
+      // [MQTT-3.1.2-11] if Will Flag is 0, Will QoS and Will Retain MUST be 0.
+      // (Will Topic / Will Message absence is already enforced by the wire format:
+      // the Netty decoder only reads them from the payload when Will Flag is 1.)
       willInvariantViolated = msg.variableHeader().willQos() != 0
-        || msg.variableHeader().isWillRetain()
-        || (willTopic != null && !willTopic.isEmpty())
-        || msg.payload().willMessageInBytes() != null;
+        || msg.variableHeader().isWillRetain();
     }
     if (willInvariantViolated) {
       endpoint = new MqttEndpointImpl(so, null, null, null, false,
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -224,6 +224,29 @@ public class MqttServerConnection {
       return;
     }
 
+    // Validate Will fields per MQTT spec section 3.1.2.5
+    boolean willInvariantViolated;
+    if (msg.variableHeader().isWillFlag()) {
+      // [MQTT-3.1.2-9] if Will Flag is 1, Will Topic MUST be present in the Payload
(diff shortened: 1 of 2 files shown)


ppatierno:
> I'm going to implement a check implementing the client MQTTv5 that verify the presente of WillTopic is WillFlag is true

So what does this PR do exactly? (together with all the same others on the branches)
Is it not complete? It should be in Draft then. It's not clear to me tbh.

dometec:
I want to fix a test that pass (I think) only becouse the client call the server implemented in this lib.
Setting WillFlag true without WillTopic cause a connection error if you try to connect to some other broker (I use VerneMQ).

ppatierno:
I got your point so I think that you should first provide a fix for the server side to return an error to the client when will flag is set but will topic is missing. It would make the current test failing which would be a valid use case. Then you could add another test with both will flag and topic which is not going to fail.

dometec:
Sure, just updated with the server side check. Let me know if I can update the other two PRs as well.

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -224,6 +224,29 @@ private void handleConnect(MqttConnectMessage msg) {
...
+    // Validate Will fields per MQTT spec section 3.1.2.5
+    boolean willInvariantViolated;
+    if (msg.variableHeader().isWillFlag()) {
+      // [MQTT-3.1.2-9] if Will Flag is 1, Will Topic MUST be present in the Payload
+      String willTopic = msg.payload().willTopic();
+      willInvariantViolated = willTopic == null || willTopic.isEmpty();
+    } else {
+      // [MQTT-3.1.2-11] if Will Flag is 0, Will QoS and Will Retain MUST be 0,
+      // and Will Topic / Will Message MUST NOT be present in the Payload
+      String willTopic = msg.payload().willTopic();
```
ppatierno: I guess you can extract the `willTopic` just one time before the if condition.

src/test/java/io/vertx/mqtt/test/server/MqttServerWillTest.java
```diff
@@ -105,6 +106,33 @@ public void testWill(TestContext context) {
...
+      client.connect(MQTT_SERVER_PORT, MQTT_SERVER_HOST).onComplete(ar -> {
+        if (ar.succeeded()) {
+          context.fail("connection should be rejected");
+        } else {
+          context.assertTrue(ar.cause().getMessage().contains("CONNECTION_REFUSED_PROTOCOL_ERROR"));
+        }
+        async.complete();
+      });
+    }));
+  }
```
ppatierno, line 181: should we also add a test for the other server condition about :

> // [MQTT-3.1.2-11] if Will Flag is 0, Will QoS and Will Retain MUST be 0,
      // and Will Topic / Will Message MUST NOT be present in the Payload

dometec: I keep only the Will QoS / Will Retain checks. The previous Will Topic / Will Message checks for the Will Flag = 0 branch were dead code, Netty's decoder only reads those payload fields when the flag bit is set, so they are always null here.

src/test/java/io/vertx/mqtt/test/server/MqttServerWillTest.java
```diff
@@ -105,6 +106,80 @@ public void testWill(TestContext context) {
...
+
+  /*
+      [MQTT-3.1.2-11] if Will Flag is 0, Will QoS and Will Retain MUST be 0.
+      (Will Topic / Will Message absence is enforced by the wire format itself,
+      so only Will QoS and Will Retain need explicit validation.)
+  */
+  @Test
+  public void testWillParamsAbsentIfWillFlagFalse1(TestContext context) {
```
ppatierno: what about `testWillParamsAbsentIfWillFlagFalseWithRetain`?

src/test/java/io/vertx/mqtt/test/server/MqttServerWillTest.java
```diff
@@ -105,6 +106,80 @@ public void testWill(TestContext context) {
...
+        } else {
+          context.assertTrue(ar.cause().getMessage().contains("CONNECTION_REFUSED_PROTOCOL_ERROR"));
+        }
+        async.complete();
+      });
+    }));
+  }
+
+  @Test
+  public void testWillParamsAbsentIfWillFlagFalse2(TestContext context) {
```
ppatierno: what about `testWillParamsAbsentIfWillFlagFalseWithQoS`?

ppatierno (approved):
LGTM. Thanks!

dometec:
Ok, I'll update the other PRs. Thanks!

PR #286, opened 2026-05-18
Add missing publish/publishRelease method that can use an identifier as argument.

vietj:
Motivation:

For QoS > 0, the client needs to re-send PUB/PUBREL messages with an existing ID.

The client is currently only capable of sending PUB/PUBREL messages with the autoassigned identifier.

When a message expires, the sender should be capable to resend the same message specifying the ID since the message is a retransmission.

Changes:

Overload MqttClient#publish MqttClient#publishRelease methods with a message identifier to implement PUB/PUBREL retransmission by the client.


ozangunalp (approved):
Thanks @vietj !

ppatierno:
@vietj I added myself as reviewer because I was going to spend some time to review it, why did you merge it? :-(

PR #288, opened 2026-05-19
Fix MqttEndpoint auth for partial CONNECT credentials

### LINKED ISSUE #287: MqttEndpoint.auth() is null for CONNECT packets with only username or only password
### Version

4.x, 5.x

### Context

`MqttEndpoint.auth()` is currently populated only when a CONNECT packet contains both the User Name and Password fields.

In `MqttServerConnection.handleConnect`, `MqttAuth` is created only when both flags are set:

```java
msg.variableHeader().hasUserName() && msg.variableHeader().hasPassword()
```

This means `endpoint.auth()` is null for username-only CONNECT packets, which are valid in MQTT 3.1.1 and MQTT 5.0. It is also null for password-only CONNECT packets, which are valid in MQTT 5.0.

Expected behavior:

- no username, no password: `endpoint.auth() == null`
- username without password: `endpoint.auth()` exposes the username with a null password
- username with password: existing behavior
- MQTT 5.0 password without username: `endpoint.auth()` exposes the password with a null username

Actual behavior:

- `endpoint.auth()` is null unless both username and password are present.

### Steps to reproduce

1. Start a Vert.x MQTT server and inspect `endpoint.auth()` in the endpoint handler.
2. Send a CONNECT packet with `User Name Flag = 1`, `Password Flag = 0`, and username `alice`.
3. `endpoint.auth()` is `null`, although the username was provided.

For MQTT 5.0, the same can be checked with `User Name Flag = 0`, `Password Flag = 1`, and a password value.

### Do you have a reproducer?

_No response_

shblue21:
### Motivation

Fixes #287.

`MqttEndpoint.auth()` was only populated when a CONNECT packet contained both username and password. As a result, username-only CONNECT packets, and MQTT 5 password-only CONNECT packets, were exposed as `null`.

### Modifications

- Create `MqttAuth` when a username is present, or when an MQTT 5 password is present.
- Leave the missing username or password value as `null`.
- Document that `MqttAuth` fields may be `null`.
- Add server connection tests for username-only and MQTT 5 password-only CONNECT packets.

### Result

`MqttEndpoint.auth()` now reflects the username/password fields that were present in the CONNECT packet.

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -257,11 +257,19 @@ public class MqttServerConnection {
         msg.payload().willProperties());
 
     // retrieve authorization information from CONNECT message
-    MqttAuth auth = (msg.variableHeader().hasUserName() &&
-      msg.variableHeader().hasPassword()) ?
-      new MqttAuth(
-        msg.payload().userName(),
-        new String(msg.payload().passwordInBytes(), CharsetUtil.UTF_8)) : null;
+    boolean hasUserName = msg.variableHeader().hasUserName();
+    boolean hasPassword = msg.variableHeader().hasPassword();
+    boolean isMqtt5 = msg.variableHeader().version() == MqttVersion.MQTT_5.protocolLevel();
+    MqttAuth auth = null;
+    if (hasUserName || (isMqtt5 && hasPassword)) {
+      String username = hasUserName ? msg.payload().userName() : null;
+      String password = null;
+      if (hasPassword) {
+        byte[] passwordInBytes = msg.payload().passwordInBytes();
+        password = new String(passwordInBytes, CharsetUtil.UTF_8);
+      }
+      auth = new MqttAuth(username, password);
+    }
 
     // check if remote MQTT client didn't specify a client-id
     boolean isZeroBytes = (msg.payload().clientIdentifier() == null) ||
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
@@ -257,18 +257,14 @@ public class MqttServerConnection {
         msg.payload().willProperties());
 
     // retrieve authorization information from CONNECT message
-    boolean hasUserName = msg.variableHeader().hasUserName();
+    boolean hasUsername = msg.variableHeader().hasUserName();
     boolean hasPassword = msg.variableHeader().hasPassword();
-    boolean isMqtt5 = msg.variableHeader().version() == MqttVersion.MQTT_5.protocolLevel();
     MqttAuth auth = null;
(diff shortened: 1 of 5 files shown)


src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -257,11 +257,21 @@ private void handleConnect(MqttConnectMessage msg) {
...
+    boolean hasUserName = msg.variableHeader().hasUserName();
+    boolean hasPassword = msg.variableHeader().hasPassword();
+    boolean isMqtt5 = msg.variableHeader().version() == MqttVersion.MQTT_5.protocolLevel();
+    MqttAuth auth = null;
+    if (hasUserName || (isMqtt5 && hasPassword)) {
+      String username = hasUserName ? msg.payload().userName() : null;
+      String password = null;
+      if (hasPassword) {
+        byte[] passwordInBytes = msg.payload().passwordInBytes();
+        if (passwordInBytes != null) {
```
ppatierno: going back through the code, it seems that when `hasPassoword` is true from the ` msg.variableHeader().hasPassword();`, the check on the password in bytes is redundant, it should not be null. Could you verify or did you find any evidence of the opposite?

shblue21: I checked the Netty decoder path: when `hasPassword` is true, `passwordInBytes()` is non-null, even for an empty password. I’ll remove the redundant check.

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
```diff
@@ -257,11 +257,19 @@ private void handleConnect(MqttConnectMessage msg) {
...
-        new String(msg.payload().passwordInBytes(), CharsetUtil.UTF_8)) : null;
+    boolean hasUserName = msg.variableHeader().hasUserName();
+    boolean hasPassword = msg.variableHeader().hasPassword();
+    boolean isMqtt5 = msg.variableHeader().version() == MqttVersion.MQTT_5.protocolLevel();
+    MqttAuth auth = null;
+    if (hasUserName || (isMqtt5 && hasPassword)) {
+      String username = hasUserName ? msg.payload().userName() : null;
+      String password = null;
+      if (hasPassword) {
+        byte[] passwordInBytes = msg.payload().passwordInBytes();
```
ppatierno: at this point this variable is redundant.

ppatierno (changes requested):

src/main/java/io/vertx/mqtt/impl/MqttServerConnection.java
shblue21: Done, removed the redundant local variables.

ppatierno (approved):
LGTM.Thanks! Would you mind to open similar PRs for the 4.x and 5.0 branches please?

shblue21:
Thanks, I'll open separate PRs for 4.x and 5.0.

PR #300, opened 2026-06-10
Fix MQTT WebSocket query string handling

### LINKED ISSUE #221: Allow Websocket URLs with Query Strings and make query string accessible for validations (extension to #208)
#### Describe the feature

With #208 it is now possible to access the http headers and the URI that was used by MQTT clients to connect to the MQTT Broker. But unfortunately it seems access to the URI is not of much use for the endpoint handler because the only allowed base address of the URI that the Server will accept correctly is "/mqtt" (without any query parameters).

Would it be possible to allow at least /mqtt?key=value style URLs to end up correctly in the EndpointHandler so that authenthentication parameters can be given with the URI to decide inside the endpoint handler if a connection attempt is legitimate?

I was unsure if this should already work (and this ticket should be a bug) or if it is a new feature.

#### Use cases

Authentication of clients based on a URL query parameter (concret mqtt url used by the client: ws://[hostname]:[port]/mqtt?auth=<AUTHENTICATION_TOKEN>

The reason why I would like to be able to access the authentication token inside the MQTT Endpoint Handler is that I want to be sure that the authentication token in the URL (which potentially is validated upstream by a reverse proxy) matches the given mqtt username/clientID/pw combination which I can only validate inside the MQTT Endpoint Handler - so that I can be sure that MQTT topic permissions match the identity of the WS connection.

shblue21:
Motivation:

Fixes #221.

MQTT over WebSocket currently accepts only the exact `/mqtt` request URI, so clients using a query string such as `ws://host:port/mqtt?auth=token` fail the WebSocket handshake.

This change accepts `/mqtt` with an optional query string while preserving the original HTTP request URI in `MqttEndpoint#httpRequestURI()`.

Non-matching WebSocket paths are still rejected before reaching the MQTT endpoint handler.

Regression tests and the WebSocket documentation were updated.

Conformance:

I have signed the ECA

### DIFF
--- a/src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
+++ b/src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
@@ -18,14 +18,22 @@ package io.vertx.mqtt.impl;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.handler.codec.MessageToMessageDecoder;
 import io.netty.handler.codec.MessageToMessageEncoder;
 import io.netty.handler.codec.compression.ZlibCodecFactory;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.FullHttpResponse;
 import io.netty.handler.codec.http.HttpObjectAggregator;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpServerCodec;
+import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.websocketx.BinaryWebSocketFrame;
+import io.netty.handler.codec.http.websocketx.WebSocketServerProtocolConfig;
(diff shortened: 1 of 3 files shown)


shblue21:
I opened this against 4.x because the issue is in the 4.5.29 milestone.

Looks like the same fix is needed on master / 5.x too. I can open the master PR first and leave this as the 4.x backport if you'd prefer.

src/main/java/io/vertx/mqtt/impl/MqttServerImpl.java
```diff
@@ -194,6 +203,29 @@ protected void encode(ChannelHandlerContext chc, ByteBuf bb, List<Object> out) t
...
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+      if (msg instanceof FullHttpRequest) {
+        FullHttpRequest request = (FullHttpRequest) msg;
+        if (!isWebSocketPath(request.uri())) {
+          FullHttpResponse response = new DefaultFullHttpResponse(
+            request.protocolVersion(), HttpResponseStatus.NOT_FOUND);
+          HttpUtil.setContentLength(response, 0);
+          ctx.writeAndFlush(response).addListener(ChannelFutureListener.CLOSE);
+          ReferenceCountUtil.release(request);
+          return;
```
vietj, line 218: this handler should remove itself after it has made his effect

shblue21: Fixed. The path handler now removes itself once the initial request is validated.

vietj:
@shblue21 can you provide PR for 5.1 and master branches as you suggested ?

PR #310, opened 2026-09-04
Fix Mqtt5ServerRedirectTest timeout by letting vertx.close() cascade to servers

tsegismont:
The compose chain bound the final future to server1's context, which was already shut down when vertx.close() completed, so the onComplete callback never dispatched. Calling vertx.close() directly avoids this and automatically closes all registered resources.


src/test/java/io/vertx/mqtt/test/client/Mqtt5ServerRedirectTest.java
```diff
@@ -60,10 +60,7 @@ public void before() {
 
   @After
   public void after(TestContext ctx) {
-    server1.close()
-      .compose(v -> server2.close())
-      .compose(v -> vertx.close())
-      .onComplete(ctx.asyncAssertSuccess());
+    vertx.close().onComplete(ctx.asyncAssertSuccess());
```
vietj, line 63: vertx.close().await() :-)

